PR #12, opened 2026-05-16
[AI assisted] Plans A + C: capabilitiesHash sealing for the Trust Card

scourtney-godaddy:
Seals the Trust Card's capabilities map into a hash that DNS records and Transparency Log entries can carry. A verifier can detect a tampered or stale capabilities list without fetching the full Trust Card JSON.

---

The V2 register endpoint accepts an optional `agentCardContent` body
holding the operator's Trust Card. The RA hashes those bytes through
JCS canonicalization (a deterministic JSON byte form per RFC 8785),
stores the SHA-256 digest on the registration row, and copies it
into the `AGENT_REGISTERED` Transparency Log event under
`attestations.metadataHashes.capabilitiesHash`. A verifier holding
the TL receipt and a fetched Trust Card can confirm the body matches
what the RA sealed, without contacting the RA itself.

The corresponding Consolidated Approach SVCB record's `card-sha256`
SvcParam carries the same digest re-encoded as base64url. The SVCB
emission ships in a follow-up PR in the stack so this PR lands the
seal independent of the DNS record format.

The `agentCardContent` field is `json.RawMessage` to keep the
operator's bytes intact through canonicalization. Round-tripping
through `map[string]any` would reorder JSON keys or normalize
numbers and shift the resulting digest. The service hashes the
bytes through the project's `anscrypto.Canonicalize` helper, stores
hex-lowercase on the aggregate, then discards the body. Empty
content leaves the digest field empty; the `AGENT_REGISTERED` event
omits the `capabilitiesHash` key entirely (`json:",omitempty"`).

Migration 006 adds `capabilities_hash` (nullable) to
`agent_registrations`. Rows registered before the column existed
load with the field empty; the activation flow does not
regenerate.

End-to-end test in `registration_capabilitieshash_e2e_test.go`
drives register → verify-acme → verify-dns → `AGENT_REGISTERED`
and confirms the sealed digest matches `SHA-256(JCS(submitted
bytes))`. Negative tests in `registration_errors_test.go` cover
malformed JSON (`INVALID_AGENT_CARD_CONTENT`) and absent-content
paths.

## Test plan

- [x] `make check` (gofmt + golangci-lint + 90% coverage gate)
- [x] End-to-end seal test (`TestRegistrationCapabilitiesHash_E2E`)
- [x] Negative path tests (malformed JSON, absent content)
- [ ] Reviewer manually verifies the migration applies cleanly against the dev DB

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #13, opened 2026-05-16
[AI assisted] Plan D: Consolidated Approach SVCB + dnsRecordStyle

scourtney-godaddy:
Adds the Consolidated Approach SVCB record format and a `dnsRecordStyle` field on the register request, so an operator can choose between the legacy four-record set and the single-record SVCB style described in the DNS-AID consensus discussion.

---

The V2 register endpoint accepts an optional `dnsRecordStyle` field
selecting which DNS records the RA emits in `dnsRecordsProvisioned`
and surfaces in the registration response for the operator to publish.
Three values:

- `consolidated` (default, recommended): one SVCB record per protocol
  endpoint at the agent's bare FQDN, carrying connection hints and
  capability locators in a single DNS lookup. Aligns with the
  Consolidated Approach DNSAID consensus discussion at
  github.com/jmozley-infoblox/DNS_for_agents_collaboration.
- `legacy`: the original `_ans` TXT family the RA has emitted since
  v0.1.x, plus an HTTPS RR (RFC 9460 type 65) at the agent FQDN
  carrying ALPN signalling for clients that don't speak ANS protocol
  detection.
- `both`: the union of consolidated and legacy, the transition shape
  for AHPs migrating from one to the other.

Empty or missing `dnsRecordStyle` normalizes to `consolidated`. An
invalid value surfaces as `INVALID_DNS_RECORD_STYLE` (HTTP 422).

The SVCB row at the bare FQDN follows RFC 9460 ServiceMode
(SvcPriority 1, TargetName `.`) so address resolution stays at the
agent's FQDN. SvcParams emitted: `alpn`, `port`, `wk` (well-known
metadata path), `card-sha256` (the base64url-encoded
`capabilitiesHash` from PR #12 in this stack). SvcParams from
DNS-AID, ANS, and other agentic specs coexist in the same record
per RFC 9460 §8 unknown-key ignore semantics.

Migration 007 adds `dns_record_style` (nullable) to
`agent_registrations`. Pre-Plan-D rows load with the field empty;
`ComputeRequiredDNSRecords` treats empty as `consolidated`.

The DNS verifier (`internal/adapter/dns/lookup.go:verifySVCB`)
mirrors the existing HTTPS-RR verifier, comparing the resolver's
parsed record against the RA's expected SvcParam set after
normalization. Quoted vs unquoted SvcParamValue forms normalize
to the same equality check; provisional SvcParamKey names that
the resolver-side formatter doesn't yet recognize surface as
non-blocking integrity findings (`Required=false`).

Tests cover SVCB emission for each style, the cross-record
hash consistency check (SVCB `card-sha256` matches the TL's
`capabilitiesHash`), and the verifier round-trip against the
miekg/dns resolver.

Stacks on #12 (Plans A + C). Merge order: #12 → this.

## Test plan

- [x] `make check` (gofmt + golangci-lint + 90% coverage gate)
- [x] SVCB emission tests for consolidated, legacy, both styles
- [x] Cross-record hash consistency test (SVCB `card-sha256` ↔ TL `capabilitiesHash`)
- [x] DNS verifier round-trip test against miekg/dns
- [ ] Reviewer confirms migration 007 applies cleanly against the dev DB


kperry-godaddy:
I opened PR #25 to cover these features.  It includes original @scourtney-godaddy commits, but re-signed by me and then some scaffolding layered on top.  Closing this for now.  We can reopen if needed.

PR #14, opened 2026-05-16
[AI assisted] Plan F: base-only registration (no version, no Identity CSR)

scourtney-godaddy:
Lets an agent register without a `version` and without an Identity CSR. The agent has no ANSName and no Identity Certificate; identity comes from the FQDN alone. Two callers want this: single-purpose agents that do not version their capabilities, and (after Plan G) anchors whose URI SAN is not FQDN-shaped.

---

ANS_SPEC.md §3.2.0 admits a base-only registration path: the
operator submits no `version` and no `identityCsrPEM`. The
resulting registration has no ANSName and the RA issues no
Identity Certificate. The agent is identified by its FQDN alone,
carried in the `agentHost` field on the `AgentRegistration`
aggregate. This PR implements the base-only path end-to-end.

Domain layer:
- `AgentRegistration` gains an `AgentHost` field (always populated
  post-validation). For versioned registrations it derives from
  `AnsName.FQDN()`; for base-only it carries the operator-supplied
  FQDN directly.
- `NewRegistration` enforces the both-or-neither invariant on
  `version` + `identityCsr`. A versioned registration without a
  CSR returns `VERSIONED_REQUIRES_IDENTITY_CSR`. A base-only
  registration with a CSR returns `BASE_ONLY_REJECTS_IDENTITY_CSR`.
- DNS records emitted for base-only registrations omit the
  `version=` field per ANS_SPEC.md §4.4.1.

Service + handler layer:
- The V2 register handler validates the both-or-neither rule before
  building the `RegisterRequest` and routes the FQDN identity to
  the service through the explicit `AgentHost` field.
- Identity-CSR signing in `VerifyACME` is gated on
  `reg.IsBaseOnly()`. Base-only registrations skip the cert
  issuance branch and the cert persistence branch; the lifecycle
  state machine still advances PENDING_VALIDATION →
  PENDING_DNS → ACTIVE through the same calls.

Storage layer:
- Migration 008 makes `ans_name` nullable. Two empty-string
  base-only rows previously collided on the UNIQUE constraint
  with code 2067. SQLite treats each NULL as distinct under
  UNIQUE, so multiple base-only registrations coexist.
- `ExistsByAnsName` short-circuits to false for the zero-value
  AnsName.
- New `ExistsActiveBaseOnlyByAgentHost` enforces FQDN uniqueness
  for base-only registrations: only one live row per FQDN.
- The V2 list and detail handlers read `AgentHost` and gate
  `version` emission on `!reg.IsBaseOnly()` so base-only rows
  emit `version: ""` rather than `"0.0.0"`.

`AnsName.String()` returns empty string for the zero value rather
than the malformed `"ans://v0.0.0."` the previous shape produced.

End-to-end behavior verified against the local demo RA: register
a base-only agent, drive verify-acme → PENDING_DNS, drive
verify-dns → ACTIVE, list and detail responses surface the
agent with `agentHost` populated and `version`, `ansName` empty.

Stacks on #12 (Plans A + C) → #13 (Plan D). Merge order:
#12 → #13 → this.

## Test plan

- [x] `make check` (gofmt + golangci-lint + 90% coverage gate)
- [x] Domain invariant tests (versioned/CSR both-or-neither)
- [x] Storage round-trip test (anchor_type, agent_host, NULL ans_name)
- [x] V2 list/detail tests (base-only emission shape)
- [x] Lifecycle test (`TestVerifyACME_BaseOnly_NoIdentityCSRRequired`)
- [x] Lifecycle regression test (`TestVerifyACME_Versioned_StillSignsIdentityCSR`)
- [ ] Reviewer confirms migration 008 applies cleanly (SQLite rebuild ceremony)

### DIFF
--- a/internal/adapter/dns/dns_test.go
+++ b/internal/adapter/dns/dns_test.go
@@ -252,6 +252,86 @@ func TestLookupVerifier_HTTPSMatch(t *testing.T) {
 	}
 }
 
+// TestLookupVerifier_SVCBMatch covers the Consolidated Approach SVCB
+// record at the bare agent FQDN. The expected value is the same
+// presentation form the RA's ComputeRequiredDNSRecords emits (see
+// internal/domain/dnsrecords.go), and the verifier matches after
+// whitespace normalization mirroring verifyHTTPS.
+//
+// Restricted to IANA-registered SvcParamKeys (alpn + port) because the
+// miekg/dns zone-file parser used by the test fixture rejects symbolic
+// names for the still-provisional Consolidated Approach SvcParams (`wk`,
+// `card-sha256`, `cap`, etc.). Until those keys are IANA-registered per
+// RFC 9460 §6, the test exercises the verifier dispatch and matching
+// path with registered keys; the unregistered keys are unit-tested at
+// the domain layer (internal/domain/dnsrecords_test.go).
+func TestLookupVerifier_SVCBMatch(t *testing.T) {
+	t.Parallel()
+	s := newTestServer(t)
+	s.add("agent.example.com.", "SVCB",
+		`agent.example.com. 3600 IN SVCB 1 . alpn=a2a port=443`)
+
+	recs := []domain.ExpectedDNSRecord{{
+		Name:     "agent.example.com",
+		Type:     domain.DNSRecordSVCB,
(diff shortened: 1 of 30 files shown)


kperry-godaddy:
@jhateley-godaddy covered this PR.

PR #18, opened 2026-05-17
[AI assisted] feat(lei): GLEIF API client + AttestationJWKSource composition

scourtney-godaddy:
Adds a real GLEIF Public API client and an AttestationJWKSource composition to the LEI anchor profile. Before this change, the LEI resolver validated mod-97 and returned `LEI_GLEIF_NOT_CONFIGURED` for every Resolve call. After this change, the resolver does end-to-end LEI verification against the GLEIF source of truth and produces a complete IdentityClaim when the deployment configures an attestation-key source.

---

## What's in this PR

Two new files plus resolver wiring.

- `internal/adapter/anchor/lei/gleif_client.go`: HTTP client against `api.gleif.org/api/v1/lei-records/{lei}`. Decodes the JSON:API Level 1 envelope (lei, legalName, status, jurisdiction, registration.lastUpdateDate). Returns nil with no error on 404 so the resolver can surface a typed "unknown LEI" failure. `WithBaseURL` for httptest; `WithHTTPClient` for production deployments that need a corporate egress proxy.

- `internal/adapter/anchor/lei/attestation_source.go`: Separate pluggable surface for the entity's attestation JWK. The GLEIF Level 1 record does not carry an attestation key; vLEI Option A and Option B both require additional deployment-specific infrastructure. `AttestationJWKSource` interface plus an in-memory `StaticAttestationSource` for testbeds and preconfigured deployments. Production deployments replace the static source with a vLEI-aware one.

- `internal/adapter/anchor/lei/resolver.go`: Resolve checks the GLEIF record's `AttestationJWK` first (vLEI-aware clients populate it inline); falls back to the source when empty. New `WithAttestationSource(s)` builder. New error code `LEI_ATTESTATION_LOOKUP_FAILED`.

## Why two surfaces

Entity verification (the GLEIF API) and attestation-key retrieval (vLEI / LOU custom field / static config) live on different infrastructure with different update cadences. Splitting them into two interfaces lets a deployment use the basic GLEIF Level 1 API for entity status while wiring its own key source for the attestation, without forcing every deployment to acquire vLEI credentials.

## Test plan

- [x] `gleif_client_test.go`: 7 tests covering happy path, 404, 5xx, bad JSON, missing required fields, context cancellation, base-URL trailing-slash handling.
- [x] `attestation_source_test.go`: 9 tests covering Set/Lookup, canonicalization on Set, defensive copies on both sides, overwrite, and three resolver-level integration cases (source supplies key, record's AttestationJWK takes precedence, no-source-no-key still errors).
- [x] `go test ./...` green across the full repo.
- [ ] Reviewer confirms the GLEIF API path matches the canonical URL the deployment will hit.

### DIFF
--- a/internal/adapter/anchor/lei/attestation_source.go
+++ b/internal/adapter/anchor/lei/attestation_source.go
@@ -1,0 +1,88 @@
+// attestation_source.go separates the entity-verification responsibility
+// (handled by GLEIFClient) from the attestation-key responsibility
+// (the entity's published verification key the agent uses to sign
+// registration events).
+//
+// The GLEIF Public API does not carry an attestation key in the
+// Level 1 record. Two paths populate the key:
+//
+//   - vLEI Option A: a self-attestation issued through GLEIF's vLEI
+//     infrastructure, fetched via that infrastructure's API.
+//   - vLEI Option B: a custom field at the entity's Local Operating
+//     Unit (LOU), retrieved through the LOU's API.
+//
+// Both paths are deployment choices, not protocol-level concerns.
+// The Resolver composes a GLEIFClient (entity verification) with an
+// AttestationJWKSource (key retrieval); the two surfaces stay
+// independent so a deployment can use the basic GLEIF Level 1 API
+// for verification and a separate pluggable source for the key.
+package lei
+
+import (
+	"context"
+	"sync"
+)
+
+// AttestationJWKSource returns the entity's published verification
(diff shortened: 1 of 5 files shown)


scourtney-godaddy:
Holding per session direction; will reopen after the full set of feature PRs is built and tested locally end-to-end against the reference agent and webmesh.ai.

PR #21, opened 2026-05-17
[AI assisted] feat(witness): OpenTimestamps witness profile (4.C)

scourtney-godaddy:
**BLUF.** Implements ANS-4 witness profile 4.C against OpenTimestamps' Bitcoin-anchored calendars. A deployment running this witness publishes an OTS proof per TL checkpoint that any verifier can validate against a Bitcoin node, anchoring TL state in a system outside the TL operator's control.

## What lands

`internal/port/witness.go` defines the `Witness` interface and the `WitnessAttestation` struct that any backend returns. The interface stays narrow: a checkpoint goes in, a backend-specific proof comes out under a profile identifier the verifier uses to pick the right validation procedure. JSON-serializable so a TL audit endpoint can round-trip an attestation without losing fidelity.

`internal/adapter/witness/opentimestamps/witness.go` is the concrete OpenTimestamps backend. `Profile()` returns `"4.C-opentimestamps"`. `Attest(ctx, checkpoint)` SHA-256s the checkpoint, POSTs the digest to `https://btc.calendar.opentimestamps.org/digest` (overridable via `WithCalendarURL`), and wraps the calendar's binary OTS proof in a `WitnessAttestation`. `Upgrade(ctx, pending)` POSTs a previously-returned pending proof back to the calendar to swap its calendar commitment for a finalized Bitcoin attestation; a 404 means "no Bitcoin block yet" and returns the input unchanged so a background refresh loop can retry.

The split between `Attest` and `Upgrade` matches OpenTimestamps' own pattern: producers call `Attest` at checkpoint time and persist the pending proof; a background loop calls `Upgrade` on a schedule (typically hourly) until the proof finalizes, then replaces the persisted bytes with the upgraded form. Verifiers run the persisted bytes through the OpenTimestamps reference CLI or any SPV-aware verifier; the witness package itself does no verification.

## Tests

`internal/adapter/witness/opentimestamps/witness_test.go` covers the happy path against an `httptest.Server`, the empty-checkpoint guard, calendar 5xx, calendar empty body, context cancellation mid-fetch, the upgrade happy path, the 404 case where the calendar treats the input as still pending, upgrade 5xx, the empty-pending guard, the default calendar URL, the trailing-slash strip on base URLs, and a compile-time interface check that `*Witness` still satisfies `port.Witness`.

## Operational notes

The default calendar is the public OpenTimestamps server. A production deployment should run its own OTS calendar instance and point `WithCalendarURL` at it; the public calendar is fine for the demo stack and integration tests but should not be a dependency in a regulated path. Switching is a one-line change at construction time.

## Test plan

- [ ] `go build ./...` clean
- [ ] `go test ./internal/adapter/witness/...` clean
- [ ] Real call against `https://btc.calendar.opentimestamps.org/digest` returns a non-empty OTS blob
- [ ] `Upgrade` returns the input unchanged the first ~10 minutes (no Bitcoin block yet) and returns finalized bytes once a block lands

### DIFF
--- a/internal/adapter/witness/opentimestamps/witness.go
+++ b/internal/adapter/witness/opentimestamps/witness.go
@@ -6,9 +6,9 @@
 //
 //  1. Hash the TL checkpoint with SHA-256.
 //  2. POST the digest to a calendar (default
-//     https://btc.calendar.opentimestamps.org/digest). The calendar
-//     returns a binary OTS proof file containing pending Bitcoin
-//     attestations.
+//     https://alice.btc.calendar.opentimestamps.org/digest). The
+//     calendar returns a binary OTS proof file containing pending
+//     Bitcoin attestations.
 //  3. Wrap the bytes in a port.WitnessAttestation with profile
 //     "4.C-opentimestamps".
 //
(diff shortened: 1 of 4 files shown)


csnitker-godaddy:
I'm going to close this and move the discussion off the PR. The open questions are about direction, and those are easier to settle in a design thread than in review.

Two decisions in this PR are worth keeping regardless of where we land: it doesn't wire the witness into checkpoint publishing, and it doesn't add an attestation field to `tl.yaml`. Both are correct under our "a registered route or wire field is a committed contract" rule — nothing is frozen, so we're free to change shape.

Why defer rather than land:

- **The port isn't an abstraction yet.** `Attest(checkpoint []byte) -> {CheckpointDigest, ExternalProof, AttestedAt}` reads as generic, but it encodes OTS's verification model (binding is a SHA-256 digest you recompute and compare) and time model (a single RFC3339 instant) under generic field names. A C2SP/Sigsum cosignature or a Hedera receipt doesn't bind or timestamp that way, so admitting a second backend forces the contract to change — which means it isn't the boundary yet.
- **Idempotency** — the checkpoint-persist path runs per append, so a naive integration attests per event, not per checkpoint.
- **Coordination** — Tessera surfaces no "I published this checkpoint" signal (the awaiter is a read-side poll), so in a replicated TL nothing tells one instance it owns witnessing.

These touch the TL service, storage, possibly a separate process, and potentially the verifier-facing contract, so I'd rather pick a direction first.

PR #23, opened 2026-05-22
fix(domain): point _ans-badge DNS record URL at transparency log

fobispo-tc:
## Summary

The `_ans-badge` TXT record's `url=` field was incorrectly set to the agent's own endpoint URL instead of the Transparency Log badge endpoint.

**Before (broken):**
```
_ans-badge.agent.example.com TXT "v=ans-badge1; version=1.0.0; url=https://agent.example.com/mcp"
```

**After (correct):**
```
_ans-badge.agent.example.com TXT "v=ans-badge1; version=1.0.0; url=https://tl.example.org/v1/agents/<agentId>"
```

## Problem

Badge-verifying clients (the SDK's `verify` package) parse the `_ans-badge` TXT record and fetch the badge from the `url=` field. The SDK documents this field as *"the URL to fetch the badge from the transparency log"* (`ans-sdk-go/verify/badge_record.go:34`), but the RA was populating it with the agent's own endpoint URL from `reg.Endpoints[0].AgentURL`. This meant verifiers would try to fetch a badge from the agent itself — which doesn't serve badges — instead of from the TL.

## Fix

- Add `tl-client.public-base-url` config field to the RA. This is the externally-reachable TL URL (e.g. `https://tl.example.org`), which may differ from the internal `tl-client.base-url` (e.g. `http://ans-tl:18081` in Docker Compose).
- `ComputeRequiredDNSRecords` now accepts the TL public URL and constructs the badge URL as `<public-base-url>/v1/agents/<agentId>`.
- Falls back to the agent endpoint URL when `public-base-url` is unset, preserving backwards compatibility for deployments that haven't configured it yet.

## Changes

| File | Change |
|------|--------|
| `internal/config/config.go` | Add `PublicBaseURL` to `TLClient` struct; default to `BaseURL` in `Validate()` |
| `internal/domain/dnsrecords.go` | Accept `tlPublicBaseURL` param; use it for badge URL when set |
| `internal/domain/dnsrecords_test.go` | Update existing tests + 2 new tests for TL URL and fallback |
| `internal/ra/service/registration.go` | Add `tlPublicBaseURL` field + `WithTLPublicBaseURL()` builder + getter |
| `internal/ra/service/lifecycle.go` | Pass `s.tlPublicBaseURL` at 3 call sites |
| `internal/ra/service/v1event.go` | Pass `s.tlPublicBaseURL` at 1 call site |
| `internal/ra/handler/dto.go` | Thread TL URL through `mapAgentDetails` and `buildRegistrationPendingBlock` |
| `internal/ra/handler/lifecycle.go` | Pass TL URL from service getter |
| `internal/ra/handler/v1registration.go` | Thread TL URL through V1 handler chain |
| `cmd/ans-ra/main.go` | Wire `cfg.TLClient.PublicBaseURL` via `.WithTLPublicBaseURL()` |
| `config/ra-local.yaml` | Document new config option (commented out) |
| `config/ra-docker.yaml` | Document new config option (commented out) |

## Configuration

```yaml
tl-client:
  base-url: "http://ans-tl:18081"          # internal — used for RA→TL HTTP calls
  public-base-url: "https://tl.example.org" # external — used in _ans-badge DNS records
```

When `public-base-url` is omitted, it defaults to `base-url`.

## Test plan

- [x] All existing tests pass (`go test ./...` — 23 packages, 0 failures)
- [x] New test: `TestComputeRequiredDNSRecords_BadgeURLPointsToTL` — verifies badge URL uses TL endpoint
- [x] New test: `TestComputeRequiredDNSRecords_BadgeFallbackWithoutTLURL` — verifies fallback to agent URL
- [ ] Deploy with `public-base-url` set and verify new registrations produce correct `_ans-badge` records


csnitker-godaddy:
Thanks for the contribution! The changes look good to me. One blocker before merge: this repo requires GPG-signed commits on main, and the two commits on this branch aren't verified yet. Could you re-sign and force-push? The Signed-off-by: trailers you already have are fine — it's just the GPG signature that's missing.

kperry-godaddy (commented):
Agree with @csnitker-godaddy - thanks for contributing.  
Three things worth addressing before merge — inline comments below.

internal/config/config.go
```diff
@@ -139,6 +139,9 @@ type StoreSQLite struct {
 type TLClient struct {
 	// BaseURL is the TL's listen URL, e.g. "http://localhost:18081".
 	BaseURL string `koanf:"base-url"`
+	// PublicBaseURL is the TL's externally-reachable URL used in
+	// _ans-badge DNS TXT records. When empty, defaults to BaseURL.
+	PublicBaseURL string `koanf:"public-base-url"`
```
kperry-godaddy, line 146: **Doc claims defaulting that doesn't exist in code.**

The comment above this field says "When empty, defaults to BaseURL," but `RAConfig.Validate()` (lines 309-352) has no such defaulting. Operators who pull this PR without setting `public-base-url` in YAML keep emitting `_ans-badge` records pointing at the agent's own URL — the exact bug this PR is fixing.

Two clean options:
- **Required + validated** (recommended): in `Validate()`, return an error when `PublicBaseURL` is empty, and reject anything that isn't `https://` with no userinfo/query/fragment.
- **Default to `BaseURL`**: assign `c.TLClient.PublicBaseURL = c.TLClient.BaseURL` when empty, so behavior matches the doc.

Either way, the field doc and the validator need to agree. The PR description's "Falls back to the agent endpoint URL when public-base-url is unset, preserving backwards compatibility" line should also be revised — the prior behavior was a bug, not a stable contract.

internal/domain/dnsrecords.go
```diff
@@ -63,8 +71,14 @@ func ComputeRequiredDNSRecords(reg *AgentRegistration) []ExpectedDNSRecord {
 	// publishing _ans without _ans-badge would advertise an agent
 	// that fails the public discovery handshake.
 	if len(reg.Endpoints) > 0 {
+		badgeURL := reg.Endpoints[0].AgentURL
+		if tlPublicBaseURL != "" && reg.AgentID != "" {
+			if joined, err := url.JoinPath(tlPublicBaseURL, "v1", "agents", reg.AgentID); err == nil {
```
kperry-godaddy: **Silent `url.JoinPath` error swallow.**

The `err == nil` guard silently falls back to `reg.Endpoints[0].AgentURL` (the buggy URL this PR is fixing). A typo in `public-base-url` — missing scheme, embedded query string, stray whitespace — is undetectable until external badge verification fails in production.

Pair this with config-load validation: parse `PublicBaseURL` once in `Validate()` and reject anything that isn't a clean `https://host[/path]`. Once `Validate()` enforces parseability, this `JoinPath` becomes a guaranteed-success invariant — either drop the error guard, or convert it to a panic so an internal regression is caught loudly rather than papered over by emitting the wrong URL into the immutable TL.

cmd/ans-ra/main.go
```diff
@@ -166,7 +166,8 @@ func run(cfgPath string) error {
 		KeyID:      signerKeyID,
 		RaID:       cfg.Signer.RaID,
 	}).WithDNSVerifier(dnsVerifier).
-		WithServerCertificateAuthority(serverCA)
+		WithServerCertificateAuthority(serverCA).
+		WithTLPublicBaseURL(cfg.TLClient.PublicBaseURL)
```
kperry-godaddy, line 175: **No startup log of effective `PublicBaseURL`.**

The TL outbox-worker startup log at lines 276-279 emits `tlBaseURL` (internal) but not `PublicBaseURL`. At 3am when oncall is debugging "badges point to the wrong place," they cannot tell pre-PR-23 from post-PR-23-but-misconfigured without SSH-ing the box and reading YAML.

Suggested addition near config load:

```go
if cfg.TLClient.PublicBaseURL == "" {
    logger.Warn().Msg("tl-client.public-base-url is unset; _ans-badge DNS records will fall back to agent URL")
} else {
    logger.Info().
        Str("tlPublicBaseURL", cfg.TLClient.PublicBaseURL).
        Str("tlBaseURL", cfg.TLClient.BaseURL).
        Msg("transparency log endpoints configured")
}
```

Collapses MTTR for badge-verification incidents from 15-30 minutes (config inspection) to a 30-second log scan.

kperry-godaddy (dismissed):

kperry-godaddy:
@fobispo-tc - looks good here.  It looks like the linter is failing. Once we get a clean build. We will get this merged. Thanks for fixing things up so far.

kperry-godaddy (approved):

kperry-godaddy:
Thanks for the contribution! 🎉

PR #25, opened 2026-05-22
feat(ra): Adapter Style DNS Discovery Profiles

kperry-godaddy:
BEGIN_COMMIT_OVERRIDE
feat(ra): Adapter Style DNS Discovery Profiles
END_COMMIT_OVERRIDE

This pull request introduces a new implementation of the ANS-family DNS discovery style (SVCB and TXT) and associated trust records, along with comprehensive unit tests and a startup-time registry/domain coherence check. The changes ensure that DNS discovery records are generated according to the latest specifications, with robust validation and clear separation of concerns between protocol logic and service wiring.

**ANS-family DNS Discovery Implementation:**

- Added a new `SVCBStyle` implementation in `internal/adapter/discovery/ans/svcb.go` that emits SVCB records for agent endpoints, including proper SvcParam composition (alpn, port, wk, card-sha256) and always includes the `_ans-badge` TXT and TLSA trust records. This makes the SVCB style self-contained and standards-compliant.
- Introduced the `BadgeRecord` function in `internal/adapter/discovery/ans/ansbadge.go` to generate the `_ans-badge` TXT record, required for trust attestation in all ANS-family styles.

**Protocol and Helper Logic:**

- Added protocol mapping helpers in `internal/adapter/discovery/ans/protocol.go` to translate protocol enums to wire tokens and well-known path suffixes, ensuring correct record emission for different agent protocols.

**Validation and Registry Coherence:**

- Implemented a startup-time registry/domain coherence check in `cmd/ans-ra/main.go` to ensure that the set of DNS discovery styles in the registry matches those advertised as valid in the domain logic, preventing configuration drift and silent failures. [[1]](diffhunk://#diff-007e4215e50321502ab3e29a004baf1617c88e1892041c3c6716f55f06dbd2e3R162-R177) [[2]](diffhunk://#diff-007e4215e50321502ab3e29a004baf1617c88e1892041c3c6716f55f06dbd2e3R411-R444) [[3]](diffhunk://#diff-007e4215e50321502ab3e29a004baf1617c88e1892041c3c6716f55f06dbd2e3R37)

**Testing:**

- Added comprehensive unit tests for SVCB record generation, badge records, protocol helpers, and port/card-sha256 logic, covering edge cases and ensuring correctness of all new logic. [[1]](diffhunk://#diff-1dd606f07b56d43b8e17974e58de8deddb4d8a525f20c21af1383ba8c280c85eR1-R256) [[2]](diffhunk://#diff-cfd2ab731b14347f92d14b446bc04a4d9e7adaeb0a9fc002179cddad5e14de7aR1-R69) [[3]](diffhunk://#diff-829ac41cd5fbf68edeb1f67582f03fbc1b8444ce2894bbfb581256eba362403cR1-R46)

These changes collectively provide a robust, standards-aligned foundation for ANS-family DNS discovery, with strong validation and test coverage.


csnitker-godaddy (changes requested):
Two concrete blockers from the DNS-AID draft itself:

- §SvcParamKeys says experimental keys MUST use the keyNNNNN form with a mandatory= gate. The PR emits friendly mnemonics directly, which dns.NewRR rejects and which the draft itself prohibits.
- The IANA Considerations table proposes cap-sha256, not card-sha256. Even after IANA registration, the PR's name won't be among the registered keys.

Until DNS-AID is WG-adopted, the SvcParamKeys IANA-registered, and the PR's names aligned with the registered set, I don't think we can merge this in a deployable state. Happy to land the parts of the PR that don't depend on the experimental keys (HTTPS RR for the TXT family, dnsRecordStyles plumbing) — that's net-useful and doesn't have a wire-format risk.

jhateley-godaddy:
Optional - I would prefer `DNSRecordStyle` -> `DiscoveryProfile` but open to other names.

internal/adapter/discovery/ans/svcb.go
```diff
@@ -0,0 +1,151 @@
...
+
+// Records returns the SVCB rows + family trust records the SVCB style
+// needs an operator to publish.
+func (s SVCBStyle) Records(reg *domain.AgentRegistration) []domain.ExpectedDNSRecord {
+	fqdn := reg.FQDN()
+	records := make([]domain.ExpectedDNSRecord, 0, len(reg.Endpoints)+2)
+	for _, ep := range reg.Endpoints {
+		alpn := protocolToANSValue(ep.Protocol)
+		wk := wkPathFor(ep.Protocol)
+		port := svcbPortFor(ep.AgentURL)
```
bchen-godaddy: On line 69, the parsing of `alpn` is reusing the `protocolToANSValue` method. 

The implementation puts a2a/mcp/http-api directly into `alpn`. It does not emit a separate protocol SvcParam at all. The svcb.go header comment (lines 30-31) makes this explicit:

▎ alpn: protocol token (a2a / mcp / http-api), distinguishes protocols within the same RRset.

This is the DNS-AID convention (overload alpn), not the ANS 3.B spec convention (alpn=h2/h3 plus separate protocol=a2a/mcp/http). The spec at dns-3b-consolidated.md:30,34 says:

```
┌─────────────────────────┬───────────────────────────┬──────────────────────────────────────┐
│          Layer          │     ANS 3.B spec says     │      PR-25 implementation does       │
├─────────────────────────┼───────────────────────────┼──────────────────────────────────────┤
│ Record location         │ bare FQDN                 │ bare FQDN ✓                          │
├─────────────────────────┼───────────────────────────┼──────────────────────────────────────┤
│ alpn value              │ h2 / h3                   │ a2a / mcp / http-api (DNS-AID        │
│                         │                           │ convention)                          │
├─────────────────────────┼───────────────────────────┼──────────────────────────────────────┤
│ Separate protocol       │ required when             │ not emitted                          │
│ SvcParam                │ multi-protocol            │                                      │
├─────────────────────────┼───────────────────────────┼──────────────────────────────────────┤
│ wk, card-sha256         │ yes                       │ yes ✓                                │
│ SvcParams               │                           │                                      │
└─────────────────────────┴───────────────────────────┴──────────────────────────────────────┘
```

Just want to check whether this is intentional?

internal/adapter/discovery/ans/svcb.go
```diff
@@ -0,0 +1,151 @@
...
+// `wk` and `card-sha256` are not yet IANA-registered SvcParamKeys;
+// see the consolidated-draft §6 note for the keyNNNNN-form fallback
+// strict-RFC parsers may need.
+type SVCBStyle struct {
+	// tlPublicBaseURL feeds the family `_ans-badge` url= via BadgeRecord
+	// (empty falls the badge back to the agent's own endpoint URL). Set
+	// once by NewSVCBStyle; styles are immutable after wiring, so Records
+	// stays a pure function of reg.
+	tlPublicBaseURL string
+}
```
bchen-godaddy: It seems to me this `SVBStyle` would be better renamed to `ANSSVCBStyle`, as it is looking for record location at `fqdn` and trying to parse SvcParams specified in `ANS 3.b` rather than DNS-AID (except the question I raised in my previous comment)?

The reason is that, in the future if we need to support DNS-AID spec, very likely we'll need a `DNSAIDSVCBStyle` which parses different SvcParams.

- ANS-defined SvcParams: alpn, port, wk, card-sha256, protocol
- DNS-AID-defined SvcParams: cap, cap-sha256, policy, realm

bchen-godaddy: Assuming DNS-AID parity isn't a scope item here, the renaming of this `SVBStyle` is not necessary at this stage.

Resolving this comment.

bchen-godaddy (approved):

jhateley-godaddy:
LGTM

internal/ra/service/lifecycle.go
```diff
@@ -898,27 +961,64 @@ func (s *RegistrationService) verifyDNSRecords(ctx context.Context, fqdn string,
...
+		// with the expected one is a hard fail regardless of the
+		// Required flag. `r.Found` is true only when the actual
+		// matched after type-specific normalization, so
+		// `DNSSECVerified && !Found` captures "response was signed,
+		// but its content disagreed with what we issued" — the exact
+		// attack we block (an attacker rewrote a record in a signed
+		// zone). Applies to TLSA (cert binding), SVCB (per-protocol
+		// service binding, including card-sha256 commitments), and
+		// HTTPS (service binding).
+		if r.DNSSECVerified && !r.Found {
```
csnitker-godaddy: This hard-fail fires on `DNSSECVerified && !Found` and `continue`s **before** the `!Required` skip, without checking `r.Actual`. On a DNSSEC-signed zone, a record the operator simply didn't publish returns NODATA, which a validating resolver authenticates (AD=true) → `(Found=false, Actual="", DNSSECVerified=true)`. That's authenticated *absence*, not tampering, but it's coded `<TYPE>_DNSSEC_MISMATCH` and bypasses `Required`.

This strands DNSSEC operators who omit an optional record — notably the `Required=false` HTTPS RR on the **default ANS_TXT profile** that CNAME-apex zones can't publish, directly contradicting the comment just below (~L989). The Required-path classification below already makes the right distinction (`!Found && Actual=="" → MISSING`), and the TXT branch here falls through on absence — SVCB/HTTPS/TLSA should too. (CI stays green only because `ans-dns` is non-validating, so AD is never set.)

Fix — gate on a present value so absence falls through to the Required check:
```go
if r.DNSSECVerified && !r.Found && r.Actual != "" {
```
Also worth a test: authenticated absence is currently covered only for TXT, not SVCB/HTTPS/TLSA.

csnitker-godaddy (commented):
Two correctness findings from a deep review (verify-dns DNSSEC classification + metadataUrl validation). Details inline.

internal/adapter/discovery/ans/protocol.go
```diff
@@ -0,0 +1,102 @@
...
+	if err != nil || u.Scheme != "https" {
+		return ""
+	}
+	if !strings.EqualFold(strings.TrimSuffix(u.Hostname(), "."), strings.TrimSuffix(fqdn, ".")) {
+		return ""
+	}
+	if !strings.HasPrefix(u.Path, wellKnownPrefix) {
+		return ""
+	}
+	suffix := strings.TrimPrefix(u.Path, wellKnownPrefix)
```
csnitker-godaddy, line 97: `validateMetadataURL` (endpoint.go) blocks SVCB-presentation-unsafe bytes in the **raw** URL, but the well-known suffix here is taken from `u.Path`, which Go has already percent-**decoded**. So `https://fqdn/.well-known/a%20b.json` passes validation, then emits `key65409=a b.json` — the space splits the SvcParam on re-parse (`strings.Fields` in `parseSVCBValue`) and is unpublishable in any real zone, stranding the agent in PENDING_DNS (`%3B`→`;` behaves the same). The sibling `key65400=<cap>` is safe because it emits the raw URL.

Fix — apply the same blocklist/non-printable check to the decoded `u.Path` (or the derived suffix) in `validateMetadataURL`, or reject when `u.RawPath != "" && u.RawPath != u.Path`.

csnitker-godaddy (approved):

PR #31, opened 2026-05-29
docs(claude): annotate Co-Authored-By override relative to global ~/.claude/CLAUDE.md

scourtney-godaddy:
The existing prohibition on AI Co-Authored-By trailers was correct but silent about why a global `~/.claude/CLAUDE.md` session might add one anyway. One parenthetical makes the override explicit so any Claude session working across both repos applies the right rule without needing to reason about the conflict.


csnitker-godaddy (approved):

csnitker-godaddy:
@scourtney-godaddy - before this can be merged the commit needs to be amended after setting up [commit signing](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) in addition you have to include a signed off by in the commit

PR #32, opened 2026-06-02
feat(verify): add provider enumeration mode to ans-verify

nicknacnic:
## Summary

Adds `ans-verify list -provider <host>` — a client-side enumeration mode that walks the log's entry tiles, decodes V1/V2 producer envelopes, and reports every agent whose `agent.host` falls under the given provider suffix. Optional flags collapse to currently-live agents, verify each match's SCITT receipt, and bound concurrency. No server-side changes; this is purely a new CLI surface on the existing `cmd/ans-verify` binary.

Use case: an operator running a provider domain (an enterprise, a demo zone, a CTF) wants to see every agent the TL has logged under that domain without registering each `agentId` out-of-band. The reference verifier already exposes per-agent reads; this lets an offline verifier reconstruct the by-provider view from the log itself.

## What's new

**CLI:**

```
ans-verify list -provider <host> [-url <tl>] [-live=false] [-verify] [-concurrency N]
```

- `-provider` — host suffix to match. Exact-match plus strict subdomain (`x.suffix`); rejects substring spoofs like `evilsuffix.com` against `suffix.com`. Case-insensitive, trailing-dot tolerant.
- `-live` (default `true`) — collapse to one row per `ansName` keeping the latest leaf, drop `AGENT_REVOKED` and `AGENT_DEPRECATED`. Unknown event types are kept (forward-compat against future active states).
- `-verify` (default `false`) — for each match, fetch and verify the SCITT receipt.
- `-concurrency N` (default 8, clamped 1–64) — parallelism for tile fetches and verify fetches.

The existing single-agent `ans-verify [-agent] <uuid>` form is unchanged.

## Security properties

The walker fails closed on every checkable property. The threat model assumes the TL endpoint could be hostile or compromised — the verifier is the trust anchor, not the server.

1. **Checkpoint signature is verified before any tile fetch.** A new `verifiedCheckpoint` fetches `/checkpoint` (raw signed note), parses it, and verifies the C2SP ECDSA-P256 signature against `/root-keys` via `logstore.VerifyC2SPECDSA`. Without this step, a hostile TL could lie about `logSize` and hide tiles containing agents the attacker wants to omit. Pure parser is `verifyCheckpointNote`.

2. **Leaf-substitution guard.** When `-verify` is on, the walker stores the JCS-canonical envelope bytes it read from each tile. After receipt verification, it asserts `bytes.Equal(receipt.ExtractPayload(rec), m.LeafBytes)`. Mismatch is a hard per-match failure with a clear "possible leaf substitution" error. Without this, a TL could serve a forged tile (fake host claim) plus a real receipt for an unrelated agent and the receipt-only check would pass.

3. **Path-injection guard on `agentId`.** The agentId on each match comes from a TL leaf the verifier doesn't fully trust. Anything that isn't a UUID is refused before being interpolated into the receipt URL.

4. **Response-size cap.** All HTTP bodies are read through `io.LimitReader(body, 32 MiB+1)`; oversize responses are rejected, not truncated.

5. **Log-injection-safe output.** All TL-supplied strings (`ansName`, `host`, `eventType`, `agentId`) are printed via `%q`, so ANSI escapes and embedded newlines cannot spoof CLI output.

6. **External context cancellation surfaces an error.** A cancelled or timed-out parent context returns the ctx error rather than a silently truncated match list. The concurrent fetcher captures the first triggering error atomically so the user sees the root cause, not whichever tile happened to be lowest-indexed at the moment of cancel.

## Out of scope

- The pre-existing `ans-verify <uuid>` single-agent path still fetches its receipt directly without first verifying a checkpoint. That's a parallel gap worth closing in a follow-up; this PR doesn't change that path.
- `EQUIVALENCE_LINK` events (PR #20) are deliberately skipped by `extractAgentIdentity` since they carry no `agent` block. A future `-include-links` mode could surface them; not needed today.
- No new TL server route. A future TL-side index of agents-by-host is on the table for scale; this client-side walk is the right primitive while logs are small to mid-sized and lets verifiers reconstruct without trusting any server index.

## Files touched

- `cmd/ans-verify/walk.go` (new) — walker, dedup, verifier, checkpoint parser, `list` subcommand.
- `cmd/ans-verify/walk_test.go` (new) — unit coverage for every helper plus integration of walker + verifier against `httptest.Server` fixtures.
- `cmd/ans-verify/main.go` — three-line subcommand dispatch at the top of `main()`. No other changes to existing behavior.

## Test plan

- [x] `make check` passes (`fmt`, `vet`, `golangci-lint`, `test-cover`).
- [x] `go test -race ./cmd/ans-verify/...` passes.
- [x] Walker correctness across tile boundary (256 → 257 leaves, parallel fetch with out-of-order completion).
- [x] Checkpoint verification: happy path, tampered body, unknown key, malformed body (subtests for each parse failure).
- [x] Leaf-substitution catch verified with a deliberately forged tile-vs-receipt mismatch.
- [x] Path-injection guard tested by failing the test if the server is ever hit with a non-UUID agentId.
- [x] Body-size cap exercised by streaming past the limit.
- [x] External `context.Cancel` surfaces a non-nil error rather than partial results.
- [x] End-to-end run against a live local TL: registered 6 agents across two providers via `scripts/demo/start.sh` + `register.sh`, ran `list` and `list -verify`, revoked one and confirmed live-mode reduction drops it, confirmed raw mode preserves the full lifecycle. Checkpoint signature verified on every invocation, all receipts verified including leaf-substitution cross-check.


nicknacnic:
Thanks for the review — all three are good catches. Addressed in `d7c288f`.

**1. Tile-size validation (`walk.go:229`).** Agreed — the checkpoint signature binds tree shape but not the contents of any individual tile, so without an explicit length assertion a hostile TL can truncate or inflate a bundle silently. Added a guard that compares decoded entry count against the expected width: `EntryBundleWidth` (256) for a full tile, the `.p/<N>` width for a partial tile. Both directions covered by new regression tests (`TestWalkProviderAgents_RejectsWrongTileSize`, `TestWalkProviderAgents_RejectsOversizedTile`).

**2. Producer respects cancellation (`walk.go:237`).** Right — the producer was pushing all N indices through the channel regardless of cancel state. Workers were drain-only on the consumer side, so for a large log this added unnecessary channel traffic and delayed teardown after the first failure. Producer now `select`s on `<-wctx.Done()` between sends and breaks out of the loop on cancel. Workers still drain whatever's already buffered via their existing `wctx.Err()` check, so `close(jobs)` never deadlocks.

**3. Flag handling (`walk.go:635`).** Good catch — `flag.ExitOnError` calls `os.Exit(2)` before `Parse` returns, so the `err != nil` block was dead. Switched to `flag.ContinueOnError` with an explicit `fs.Usage` so parse failures print consistent usage text and exit with code 1, matching the rest of the binary.

Diff: nicknacnic:feat/ans-verify-list-provider → d7c288f

csnitker-godaddy:
Thanks for the contribution! This PR actually surfaced a bug in the checkpoint signing that differed from the intended signature format. Could you please update your PR with the latest changes from main and update the parsing to handle the correct format?

nicknacnic:
Thanks Connor — pulled in `9717c0e` (your merge of `main`) which brings PR #38 under us.

The walker's production code didn't need changes: `verifyCheckpointNote` delegates ECDSA verification to `logstore.VerifyC2SPECDSA`, and after #38 that function accepts ASN.1 DER as primary plus IEEE P1363 as a legacy fallback. So your merge alone made the walker correct against the corrected wire format.

What did need updating was the **test fixture** — `signTestCheckpoint` in `walk_test.go` was synthesizing checkpoint notes with P1363 r||s signatures (the old buggy format). It was still passing because of the P1363 fallback, but the test was pinning the deprecated shape, not the production one. Updated it to emit DER in `d0fa648`, on top of your merge.

Verified end-to-end against a fresh build of this branch: brought up `scripts/demo/start.sh`, registered three agents under `connor-test.example`, ran `ans-verify list -verify`. Output:

```
Tree size:   3 leaves (checkpoint signature ✓)
Matched:     3 live agents (from 3 raw leaves)
...
Verified 3/3 receipts (0 failed)
```

Branch tip: `d0fa648`.

csnitker-godaddy (approved):
Thanks for getting that updated! The changes look good and I verified they work against the local and production transparency log.

I think we probably need to come back to this cli sooner than later and implement something like cobra / viper to manage the CLI flags and help text but I can circle back on that when I get a bit of time

PR #35, opened 2026-06-03
feat(ra): make version optional on agent registration

jhateley-godaddy:
The `version` field on agent registration is now optional in both the V1 and V2 RA lanes. When omitted, the agent's version defaults to 1.0.0; an explicit empty string (or any malformed value) is rejected as 422 MALFORMED_SEMVER, distinguishing an absent field from a bad one.

- domain.ResolveVersion maps an optional wire version (`*string`) to a domain semver: nil → DefaultVersion (1.0.0), non-nil → strict ParseSemVer.
- Registration DTOs change Version from string to *string so JSON omission is observable.
- spec/api-spec-v2.yaml and the docsui mirror drop `version` from the required list and document the optional/default behaviour.
- Tests cover omitted (default), supplied (preserved), and empty/malformed (rejected) across both lanes.

### DIFF
--- a/internal/adapter/docsui/openapi/ra.yaml
+++ b/internal/adapter/docsui/openapi/ra.yaml
@@ -1048,6 +1048,11 @@ components:
           type: string
           pattern: '^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)$'
           example: 1.2.0
+          description: >-
+            Optional. A supplied value is validated against the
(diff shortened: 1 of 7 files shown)


csnitker-godaddy (changes requested):
IMO we should not be defaulting this value. The overhead to providing this value is very minimal and doesn't really reduce friction for the consumer.

jhateley-godaddy:
As discussed - there is little value in omitting the version from the registration payload, closing this PR.

PR #41, opened 2026-06-10
feat: verified identities — the who behind agents (did:web + did:key)

csnitker-godaddy:
## Summary

Implements **Verified Identities** — the "who" behind an agent (the "what") — per the team's `DESIGN-multi-identity-anchors.md` rev 4 design of record (who/what pivot, 2026-06-10), including the two post-design corrections: **multi-algorithm key support** and **verbatim verification-method sealing**.

An identity is a first-class object owned by the `providerId`, proven through a challenge-bound control proof, sealed onto its own Transparency Log stream, and linked to any number of that owner's agents. Prove once, link the fleet; rotation/revocation is **one sealed event** regardless of linked-agent count — propagation is the TL's read-time join. **The agent registration path is byte-for-byte untouched.**

## What's in here (4 commits, reviewable independently)

1. **feat(tl)** — the `IDENTITY_*` event family (closed 5-token enum, same JCS + RFC 6962 leaf rules), the `POST /v1/internal/identities/event` ingest lane on the existing producer-signature discipline, the identity read surface (`/v1/identities/{id}{,/audit,/receipt,/agents}`), and the agent badge's computed `identities[]` join. One Merkle tree; "streams" are read indexes (`tl_events.identity_id` + a link fan-out table).
2. **feat(ra)** — the `/v2/ans/identities` surface: register → verify-control → rotate/revoke + the owner-gated links. Per-kind logic lives behind the **`controlVerifier` registry** (`internal/ra/service/identitykinds.go`) — the extension seam for `did:plc`, `did:ion`, `did:ethr`, and `lei` (vLEI). `did:web` + `did:key` ship enabled.
3. **feat(demo)** — `scripts/demo/identity-lifecycle.sh` exercises **every** operation live (multi-key ES256+EdDSA proof, Ed25519 did:key side-by-side with the did:web on one agent, two-agent batch link sealing one event, rotation 2→1 keys, per-pair unlink, read-time revocation propagation); `run-lifecycle.sh` gains the identity steps; `signproof` is the registrant-side key/JWS tool.
4. **docs** — the implementation plan + as-built notes.

## Security invariant (stated once)

The RA seals an identity attestation **only after control is proven** — never on resolution alone. Every submitted compact JWS must carry the served JCS `IdentityProofInput` (`{identifier, identityId, nonce, purpose:"ans:identity-proof:v1", raId, scheme}`) **verbatim** as its payload, and verify against the identifier's **authoritative** key (the re-fetched `did.json` for did:web; the identifier itself for did:key). One bad proof fails the call closed. The nonce is single-use, consumed only inside the success transaction (conditional-UPDATE TOCTOU guard); failed attempts never consume it.

## Pre-implementation shape diff (per CLAUDE.md)

The canonical contracts under `spec/` are **authored in this PR** (both validate as OpenAPI 3.0.3; docsui copies synced via `make docs-sync`):

- `spec/api-spec-v2.yaml` — 8 identity routes (`/ans/identities{,/{id}}`, `verify-control`, `revoke` (a POST — identities are never deleted), `links{,/{agentId}}`) + `AgentDetails.identities[]` (additive, computed).
- `spec/api-spec-tl-v2.yaml` — the identity ingest + read surface, `IdentityProducerEvent`, `ProvenKey = {verificationMethod (verbatim), signedProof}`, `LinkedIdentityView.provenKeyIds`.

Deliberate deltas from the rev-4 design text, all discussed with Connor/team:

| Design said | This PR does | Why |
|---|---|---|
| widen the V2 `Type.IsValid()` enum | parallel `internal/tl/event/identity` package, own closed enum | agent codec stays byte-frozen; cross-lane 422 falls out of each codec's `Validate()` |
| separate `tl_identity_events` table | nullable `identity_id` column on `tl_events` + link fan-out table | the design's own maxim — streams are read indexes over one log |
| seal `keyThumbprint` + normalized `publicKeyJwk` | seal the document's `verificationMethod` **verbatim**; thumbprints compute-at-read | post-design correction (no derived values in seals); lei keeps its AID+thumbprint exception |
| (implicit single alg) | EdDSA + ES256 + RS256; precise rejections for X25519/secp256k1/P-384+ | support exactly what the JWS layer verifies |

## Extensibility (the vlei/did:ion/did:plc path)

Adding a kind = one `controlVerifier` implementation + one registry entry + a lexical arm in `domain.InferIdentifierKind`. JWS-scheme kinds reuse the shared proof machinery wholesale with a different fetcher port (noop/real adapter split, same as the DNS verifier). Per-kind wire shapes are additive members (`signedProofs` today; `cesrSignature`/`ethSignature` later — exactly one family per kind). Unregistered kinds → `IDENTIFIER_KIND_UNSUPPORTED`; unknown did methods name the method in the error. No `kind` CHECK in storage — the domain dispatcher + registry are the single source of truth.

## Verification

- `make check` green: fmt, vet, golangci-lint (0 issues), coverage **90.1%** ≥ 90% gate (`internal/domain` additions at 100%)
- `make test-race` clean (incl. concurrent nonce-consumption)
- Both demos run end-to-end against the live stack (RA + TL + outbox + Tessera): every `IDENTITY_*` token sealed, COSE receipts verify, `ans-verify` passes on the agent path
- Cross-lane guards tested in both directions; V1 lane frozen

## Notes for reviewers

- Sealed shapes are append-only-forever — **the event vocabulary in commit 1 is the thing to scrutinize hardest** (design §5.5 sequencing note; ANS-spec amendment rows A–F still need ratification upstream before a shared TL seals these).
- The noop resolver is quickstart-only (real signature verification, waived live-document binding) — config-gated exactly like `dns.type: noop`.
- Commits are DCO-signed-off; GPG signing isn't configured in the authoring environment, so signing needs to happen at merge.


kperry-godaddy (approved):
Good work here!

internal/adapter/tlclient/client.go
```diff
@@ -232,3 +240,29 @@ func IsPermanent(err error) bool {
...
+// failed delivery IS a failed operation, so errors are mapped to
+// domain kinds the service layer can return directly:
+//
+//   - transient (transport, 5xx, 429) → ErrUnavailable / 503
+//     TL_UNAVAILABLE: retryable, the caller consumed nothing;
+//   - permanent (other 4xx)           → ErrInternal /
+//     TL_REJECTED_EVENT: the RA produced an event the TL refuses —
+//     a pipeline bug, not weather; operators must see it.
+func (c *Client) SealIdentityEvent(ctx context.Context, innerCanonical []byte, producerSig string) error {
+	_, err := c.Append(ctx, "IDENTITY", innerCanonical, producerSig)
```
jhateley-godaddy, line 257: We should update the comment in `func (c *Client) Append(ctx context.Context, schemaVersion string, body []byte, producerSig string) (*AppendResult, error)`
Still says:
```
// schemaVersion must be "V1" or "V2" — it selects the URL path
// (`/v1/internal/agents/event` vs `/v2/internal/agents/event`).
```

internal/crypto/jwk.go
```diff
@@ -0,0 +1,342 @@
...
+	switch f.Crv {
+	case "Ed25519":
+		x, err := base64.RawURLEncoding.DecodeString(f.X)
+		if err != nil {
+			return nil, fmt.Errorf("%w: decode x: %w", ErrJWK, err)
+		}
+		if len(x) != ed25519.PublicKeySize {
+			return nil, fmt.Errorf("%w: Ed25519 key must be %d bytes", ErrJWK, ed25519.PublicKeySize)
+		}
+		return ed25519.PublicKey(x), nil
```
jhateley-godaddy, line 103: my IDE flags this as a redundant type conversion.  Up to you - I don't really see a problem with being explicit about the type.

internal/adapter/didresolver/web.go
```diff
@@ -0,0 +1,393 @@
...
+					"DID document redirect left https")
+			}
+			redirDomain, derr := registrableDomain(req.URL.Hostname())
+			if derr != nil || redirDomain != originDomain {
+				return domain.NewValidationError("DID_REDIRECT_DOMAIN_MISMATCH",
+					"DID document redirect left the DID's registrable domain")
+			}
+			return nil
+		},
+	}, nil
```
jhateley-godaddy, line 228: I can see that we support redirects for sub-domains but not cross domain redirects.  I am wondering if this is really necessary to support sub-domains, or even redirects - at all.

internal/ra/service/identity.go
```diff
@@ -0,0 +1,925 @@
...
+		current, err := s.identities.FindByID(txCtx, identityID)
+		if err != nil {
+			return err
+		}
+		if current.Status != domain.IdentityVerified {
+			return domain.NewInvalidStateError("IDENTITY_NOT_VERIFIED",
+				"identity was revoked while the link was sealing; the sealed link event is inert")
+		}
+		for _, agentID := range newlyLinked {
+			if _, err := s.links.Link(txCtx, identityID, agentID, now); err != nil {
```
jhateley-godaddy, line 728: If there is a transient database error here, we would have sealed the event with the agentIDs but not saved the links in the RA database. Does it matter? The TL would be authoritative and the intention was to link the agent ids.    

In the REVOKED case I don't think it's a problem because the revoke event would have sealed a later event.

internal/adapter/didresolver/web.go
```diff
@@ -0,0 +1,393 @@
...
+	return dialer.DialContext(ctx, network, net.JoinHostPort(pinnedIP, port))
+}
+
+// resolveAndPin resolves the host, applies the egress denylist to
+// every candidate address, and pins the first allowed IP. First
+// writer wins — concurrent dials for one host converge on one pin.
+func (d *pinningDialer) resolveAndPin(ctx context.Context, host string) (string, error) {
+	ips, err := net.DefaultResolver.LookupIPAddr(ctx, host)
+	if err != nil || len(ips) == 0 {
+		return "", errors.New("didresolver: host did not resolve")
```
jhateley-godaddy, line 289: It might be useful for debugging if we include the err and host here, or log it.
```
 if err != nil {                                                                                                                                                                                                
      return "", fmt.Errorf("didresolver: resolve %q: %w", host, err)                                                                                                                                            
  }                                                                                                                                                                                                              
  if len(ips) == 0 {                                                                                                                                                                                             
      return "", fmt.Errorf("didresolver: %q resolved to no addresses", host)                                                                                                                                    
  }       
```

internal/adapter/didresolver/web.go
```diff
@@ -0,0 +1,393 @@
...
+
+func (d *pinningDialer) DialContext(ctx context.Context, network, addr string) (net.Conn, error) {
+	host, port, err := net.SplitHostPort(addr)
+	if err != nil {
+		return nil, errors.New("didresolver: malformed dial address")
+	}
+	if port != "443" {
+		// did:web fetches are pinned to 443; nothing else should
+		// ever reach the dialer.
+		return nil, errors.New("didresolver: refusing non-443 connection")
```
jhateley-godaddy, line 261: can we add the `addr` here and on line 256?

jhateley-godaddy (commented):
Looks great

PR #45, opened 2026-06-13
feat(ra): pluggable server-cert issuance via certificate-order lifecycle

csnitker-godaddy:
## Summary

Wires an RFC 8555 ACME provider (Let's Encrypt–shaped) alongside the local
self-signed CA behind a single `port.ServerCertificateIssuer`, so a deployment
can swap in a public or cloud CA (AWS Private CA, GCP CAS, etc.) without
touching the service layer. Issuance now runs through an explicit
**certificate-order lifecycle** rather than assuming synchronous, local signing.

## Order lifecycle

- **`CreateOrder`** (at registration / renewal submission) returns the
  provider's domain-control challenges, relayed verbatim to the domain owner.
  ANS never writes DNS or serves challenge files on the owner's behalf — the
  pending response carries the records/paths for the owner to publish.
- **`FinalizeOrder`** (at `verify-acme`, gated on a verified challenge) returns
  the certificate, or `ErrOrderPending` for asynchronous providers such as ACME
  CAs. Re-POSTing `verify-acme` re-drives the pending order (`202` /
  `PENDING_CERTS`), so the async path is plumbed end-to-end instead of being
  skipped.

Identity certificates remain privately issued behind
`port.IdentityCertificateAuthority` (self-CA default; swappable for AWS Private
CA / GCP CAS) and are signed **only after the server order completes**. The
identity CA port gains an idempotent `RevokeCertificate`. Server-cert revocation
is intentionally out of scope on the `ServerCertificateIssuer` port for now
(ACME RFC 8555 §7.6 for public CAs, a noop for BYOC); it can be added when a
server-cert revoke flow is needed.

## Correctness / hardening

Closes holes so the flows behave correctly regardless of the underlying adapter:

- Always gate `FinalizeOrder` on a verified DNS-01 **or** HTTP-01 artifact;
  gate/order rejections return `422`, pending-conflict `409`.
- SSRF-harden the HTTP-01 verifier: guarded dialer rejecting non-public dial
  targets (loopback, link-local incl. the cloud metadata endpoint, RFC 1918,
  ULA, RFC 6598 CGNAT `100.64.0.0/10`, 6to4, NAT64), refuse redirects, bounded
  body read — fails closed.
- Distinguish "no server cert on file" from a transient store failure, so a
  recoverable fault can never activate an agent and sign a wrong, **immutable**
  transparency-log leaf.
- Persist order/challenge state and certificate serial/ref in SQLite
  (migrations `008`/`009`); guarded set-based expiry of lapsed pending
  validations; in-process agent auto-expiry sweeper.

## Wire contract

`spec/api-spec-v2.yaml` and the docsui copy stay byte-identical. `PENDING_CERTS`
is restored to `RegistrationPending.status`; `CANCEL` added to the next-step
action enum.

## Testing

- `make check` passes (`fmt`, `vet`, `golangci-lint`, coverage gate).
- `internal/domain` at 100% statement coverage; overall ≥ 90%; `make test-race`
  clean.
- New unit suites cover the order lifecycle, the ACME adapter (against an
  in-process RFC 8555 fake), the SSRF guard, and the SQLite persistence
  round-trips.
- The ACME provider is wired into `scripts/demo/start.sh` behind `--with-acme`
  for live testing against a real domain + Let's Encrypt staging.

## Review feedback addressed

- Copilot flagged that the HTTP-01 SSRF guard cloned `http.DefaultTransport`,
  inheriting `ProxyFromEnvironment` — an `HTTP_PROXY`/`HTTPS_PROXY` could route
  the fetch through a (public, allowed) proxy that then reached an internal
  target. Fixed by setting `Proxy = nil` so the transport always dials the
  resolved FQDN directly and the dial guard stays authoritative.

## Notes

Marked draft for review of the port surface before merge. `main` has been merged
in; migrations are renumbered to `008`/`009` to sequence after the verified-
identities migrations (`006`/`007`).


internal/adapter/cert/acme.go
```diff
@@ -0,0 +1,460 @@
...
+
+// ACMEIssuerOption configures the issuer at construction time.
+type ACMEIssuerOption func(*ACMEIssuer)
+
+// WithFinalizeBudget overrides how long FinalizeOrder blocks on the
+// provider before reporting port.ErrOrderPending.
+func WithFinalizeBudget(d time.Duration) ACMEIssuerOption {
+	return func(a *ACMEIssuer) { a.finalizeBudget = d }
+}
+
```
jhateley-godaddy, line 96: Let's add a withLogger - similar to `service/identity.go` L152

internal/adapter/cert/acme.go
```diff
@@ -0,0 +1,460 @@
...
+		return &domain.CertificateOrder{
+			OrderRef:   order.URI,
+			State:      domain.OrderStatePending,
+			Challenges: challenges,
+			ExpiresAt:  order.Expires,
+		}, nil
+	case acme.StatusReady, acme.StatusProcessing, acme.StatusValid:
+		// Authorizations already satisfied (reuse) or issuance already
+		// underway — no domain-control artifact for the owner to
+		// publish. Relay as ISSUING so verify-acme drives FinalizeOrder.
```
jhateley-godaddy, line 194: Can we log the orderRef and state here.

internal/adapter/cert/acme.go
```diff
@@ -0,0 +1,460 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("cert: acme new-order: %w", err)
+	}
+
+	switch order.Status {
+	case acme.StatusPending:
+		challenges, cerr := a.collectChallenges(ctx, order)
+		if cerr != nil {
+			return nil, cerr
+		}
```
jhateley-godaddy, line 179: Can we log INFO order opened {fqdn, orderRef}

internal/adapter/cert/acme.go
```diff
@@ -0,0 +1,460 @@
...
+		if err := a.answerVerifiedChallenges(ctx, order, req.Verified); err != nil {
+			return nil, err
+		}
+		if order, err = a.waitOrder(ctx, req.OrderRef); err != nil {
+			return nil, err
+		}
+	}
+
+	switch order.Status {
+	case acme.StatusReady:
```
jhateley-godaddy, line 301: INFO finalizing order {fqdn, orderRef, serialNumber}

internal/adapter/cert/acme.go
```diff
@@ -0,0 +1,460 @@
...
+	case acme.StatusValid:
+		der, err := a.client.FetchCert(ctx, order.CertURL, true)
+		if err != nil {
+			return nil, fmt.Errorf("cert: acme fetch cert: %w", err)
+		}
+		return a.issuedFromChain(der, order.CertURL)
+	case acme.StatusInvalid:
+		return nil, fmt.Errorf("cert: acme order invalid: %w", port.ErrOrderFailed)
+	default:
+		return nil, fmt.Errorf("cert: acme order in unexpected status %q", order.Status)
```
jhateley-godaddy, line 329: can we log these error scenarios - could be useful to debug upstream errors.

jhateley-godaddy (changes requested):
Looks good.  Tested demo with acme.  Was able to get an agent active with a server certificate issued from LE.  

I had some issues doing a cert renewal with a CSR I created.
The response from CSR submission looked fine:
```
{
  "renewalType": "SERVER_CSR",
  "status": "PENDING_VALIDATION",
  "csrId": "8201337a-64f2-4e49-8906-5497c5aee24d",
  "expiresAt": "2026-06-25T16:29:43+10:00",
  "nextStep": {
    "action": "VALIDATE_DOMAIN",
    "endpoint": "/v2/ans/agents/8d1e69ee-ea7c-4be0-857b-38db9262198a/certificates/server/renewal/verify-acme",
    "description": "Complete ACME challenges then POST to verify-acme endpoint"
  },
  "links": [
    {
      "rel": "status",
      "href": "/v2/ans/agents/8d1e69ee-ea7c-4be0-857b-38db9262198a/certificates/server/renewal"
    }
  ]
}
```

But when trying the verify-acme endpoint there was an issue which I couldn't diagnose - this might be a quirk of LE. 

```
 {                                                                                                                                                                                                              
    "type": "about:blank",                                                                                                                                                                                       
    "title": "Validation Failed",                                                                                                                                                                                
    "status": 422,                                                                                                                                                                                               
    "detail": "certificate provider reported a terminal order failure; submit a new renewal",                                                                                                                    
    "code": "CERT_ORDER_FAILED"                                                                                                                                                                                  
  }           
```

Let's add some more logging for the error branches - will help implementers to diagnose issues.

internal/adapter/cert/acme.go
csnitker-godaddy: Added `WithLogger(zerolog.Logger)` — component-tagged (`component=acme-issuer`), defaulting to `zerolog.Nop()`, mirroring `IdentityService.WithLogger` — and wired it through `buildServerIssuer` in `cmd/ans-ra`. Done in a924551.

internal/adapter/cert/acme.go
csnitker-godaddy: Added `INFO "acme order opened" {fqdn, orderRef, acmeStatus}` immediately after `AuthorizeOrder` succeeds. a924551.

internal/adapter/cert/acme.go
csnitker-godaddy: Each `CreateOrder` branch now logs the relayed state: DEBUG on the pending branch with the challenge count, and INFO on the authorization-reuse branch (relayed as ISSUING). Both carry the orderRef. a924551.

internal/adapter/cert/acme.go
csnitker-godaddy: Added `INFO "acme finalizing order" {fqdn, orderRef}` at the start and `INFO "acme order finalized" {fqdn, orderRef, serialNumber}` on success (serial is only known post-issuance, so it's on the completion line). a924551.

internal/adapter/cert/acme.go
csnitker-godaddy: The upstream-error scenarios now log at ERROR with `.Err(err)` + orderRef — get-order, answer-challenges, fetch-cert, and the invalid/unexpected terminal statuses. The expected `ErrOrderPending` re-drive path logs at DEBUG (not an error). a924551.

internal/adapter/cert/serverselfca.go
```diff
@@ -85,15 +116,54 @@ func NewServerSelfCA(dataDir, org string, validityDays int, opts ...ServerSelfCA
...
 	ctx context.Context,
-	csrPEM string,
-	fqdn string,
+	req port.FinalizeOrderRequest,
 ) (*port.IssuedCert, error) {
-	csr, err := anscrypto.ValidateServerCSR(csrPEM, fqdn)
+	csr, err := anscrypto.ValidateServerCSR(req.CSRPEM, req.FQDN)
 	if err != nil {
 		return nil, err
 	}
```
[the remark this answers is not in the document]
csnitker-godaddy: Good call — `ServerSelfCA.FinalizeOrder` now requires a non-empty `OrderRef` (returns an error otherwise). The stateless self-CA still doesn't look anything up by it, but this keeps the contract uniform with the ACME issuer (which needs the ref to resume the order), so a dropped ref fails the same way in dev as against Let's Encrypt. Done in a924551.

internal/adapter/challenge/http.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+	base, ok := http.DefaultTransport.(*http.Transport)
+	if !ok {
+		// The stdlib default is always *http.Transport; fall back to a
+		// fresh one if some test rewired it.
+		base = &http.Transport{}
+	}
+	t := base.Clone()
+	t.DialContext = d.DialContext
+	return t
```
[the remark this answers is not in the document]
csnitker-godaddy: Fixed in 602b309 — `guardedTransport()` now sets `Proxy = nil`, so the transport always dials the resolved FQDN directly and the dial guard stays authoritative regardless of `HTTP_PROXY`/`HTTPS_PROXY`. Added `TestSSRFGuard_DisablesProxy`.

internal/port/certauthority.go
```diff
@@ -76,45 +129,102 @@ type IdentityCertificateAuthority interface {
...
+// same way, returning no challenges from CreateOrder when they perform
+// no domain validation of their own.
+//
+// The server-cert trust root is distinct from the identity CA so the
+// two can be rotated and key-managed independently.
+type ServerCertificateIssuer interface {
+	// CreateOrder opens an issuance order for the agent's FQDN and
+	// returns the provider's domain-control challenges. The CSR is NOT
+	// required yet (matching ACME, where the CSR is presented at
+	// finalize). The returned order is persisted on the registration
```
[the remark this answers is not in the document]
csnitker-godaddy: Correct — only `IdentityCertificateAuthority` has `RevokeCertificate`; `ServerCertificateIssuer` does not. Updated the PR description to scope the claim to the identity CA port and note server-cert revocation is intentionally out of scope for now (ACME RFC 8555 §7.6 for public CAs, a noop for BYOC).

jhateley-godaddy (approved):
Looks good.

PR #46, opened 2026-06-15
feat(ra): Implement ARD-compliant discovery service and event feed

kperry-godaddy:
This pull request introduces the new `ans-finder` service, which implements the Agentic Resource Discovery (ARD) service over the ANS reference implementation. It also makes several improvements to the RA service, including the addition of a public agent-events feed and enhanced handler logging. The Makefile is updated to support building and documenting the new finder service. Minor cleanups and dependency updates are also included.

**New ANS Finder Service:**

- Added the new `ans-finder` binary (`cmd/ans-finder/main.go`), which:
  - Polls the RA's agent-events feed, indexes events for full-text search, and serves public discovery endpoints per the ARD contract.
  - Includes liveness/readiness endpoints and operational runbook documentation.
  - Features a rate-limited, anonymous HTTP API and a Swagger UI docs endpoint.

**Makefile and Build System Updates:**

- Updated the `Makefile` to support building the new `ans-finder` binary and to sync its OpenAPI spec into the docs UI. [[1]](diffhunk://#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L1-R1) [[2]](diffhunk://#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L31-R31) [[3]](diffhunk://#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R49-R52) [[4]](diffhunk://#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R61)

**RA Service Improvements:**

- Added a public `/v1/agents/events` feed to the RA, with exact-path anonymous access and handler wiring for event streaming. [[1]](diffhunk://#diff-007e4215e50321502ab3e29a004baf1617c88e1892041c3c6716f55f06dbd2e3R98) [[2]](diffhunk://#diff-007e4215e50321502ab3e29a004baf1617c88e1892041c3c6716f55f06dbd2e3R178-R197) [[3]](diffhunk://#diff-007e4215e50321502ab3e29a004baf1617c88e1892041c3c6716f55f06dbd2e3L233-R284) [[4]](diffhunk://#diff-007e4215e50321502ab3e29a004baf1617c88e1892041c3c6716f55f06dbd2e3R404-R409) [[5]](diffhunk://#diff-007e4215e50321502ab3e29a004baf1617c88e1892041c3c6716f55f06dbd2e3R420-R423)
- Enhanced security by adding an HTTP middleware to set `X-Content-Type-Options: nosniff` on all responses, especially for the public events feed.
- Updated all handler constructors to accept a logger, improving logging consistency and observability. [[1]](diffhunk://#diff-007e4215e50321502ab3e29a004baf1617c88e1892041c3c6716f55f06dbd2e3L197-R213) [[2]](diffhunk://#diff-007e4215e50321502ab3e29a004baf1617c88e1892041c3c6716f55f06dbd2e3L206-R222) [[3]](diffhunk://#diff-007e4215e50321502ab3e29a004baf1617c88e1892041c3c6716f55f06dbd2e3L233-R284)

**Code Cleanup and Dependency Updates:**

- Cleaned up unused imports and removed the unused `VerifiedCheckpoint` struct in `cmd/ans-verify/walk.go`. [[1]](diffhunk://#diff-04e4c9a10ff1ffeec2b70c7ec4956d5f24cf00ad92642298c47aaadd39d6bd6dL7-L8) [[2]](diffhunk://#diff-04e4c9a10ff1ffeec2b70c7ec4956d5f24cf00ad92642298c47aaadd39d6bd6dL18) [[3]](diffhunk://#diff-04e4c9a10ff1ffeec2b70c7ec4956d5f24cf00ad92642298c47aaadd39d6bd6dL27-R24) [[4]](diffhunk://#diff-04e4c9a10ff1ffeec2b70c7ec4956d5f24cf00ad92642298c47aaadd39d6bd6dL481-L487)


spec/api-spec-v2.yaml
```diff
@@ -71,6 +79,101 @@ security:
...
+        '200':
+          description: Successful operation
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/EventPageResponse'
+        '422':
+          description: Invalid request parameters (out-of-range or non-numeric limit)
+          content:
+            application/problem+json:
```
jhateley-godaddy, line 173: The 422/500 responses here are declared as `application/problem+json` (correct — that's what the handler emits), but the schema points at `ErrorResponse`, which is the wrong shape.                             
                                                                                                                                                                                                      
- `writeProblem` (internal/ra/handler/errors.go:90) serializes the RFC 7807 Problem struct:                                                                                                                      
  `{ "type": "about:blank", "title": "Validation Failed", "status": 422, "detail": "...", "code": "INVALID_LIMIT" }  `                                                                                           
- `ErrorResponse` (api-spec-v2.yaml:1747) declares `{ status: "ERROR" (string enum), code, message, details }` — mismatched on `status` (int vs string), detail vs message, and missing type/title.              
                                                                                                                                                                                               
- Fix: add a `Problem` component schema matching the RFC 7807 shape (type/title/status required, detail/code optional) and reference it here instead of ErrorResponse. Keep the `application/problem+json` content  type as-is.   Similar to the one defined in `api-spec-finder-v1.yaml`

spec/api-spec-v2.yaml
```diff
@@ -71,6 +79,101 @@ security:
...
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+        '500':
+          description: Internal server error
+          content:
+            application/problem+json:
```
jhateley-godaddy, line 179: same as my comment on line 167.

jhateley-godaddy (changes requested):

internal/adapter/store/sqlitefinder/search.go
```diff
@@ -0,0 +1,191 @@
...
+		}
+		if a.row.Type != b.row.Type {
+			return a.row.Type < b.row.Type
+		}
+		return a.row.URL < b.row.URL
+	})
+}
+
+// rfc3339 formats now for lexical comparison against the index's stored
+// RFC 3339 expires_at strings. The feed normalizes its timestamps so a
```
jhateley-godaddy: Nit: "the feed normalizes its timestamps" is false — feed.go:219 validates with time.Parse but discards the result and stores the raw string. The lexical compare is sound only because the RA emits canonical UTC Z second-precision; nothing enforces it. Either normalize at the projection boundary or reword to state the dependency.

jhateley-godaddy (changes requested):

spec/api-spec-v2.yaml
kperry-godaddy: Great catch, James — thank you. You're right that the schema was out of step with the content type. I've added a `Problem` component schema matching the RFC 7807 shape the handler actually emits (`type`/`title`/`status` required, `detail`/`code` optional), modeled on the one in the finder spec, and pointed both the `422` and `500` at it. `ErrorResponse` stays for the routes that still use it, and the `application/problem+json` content type is unchanged. Appreciate the close read here.

spec/api-spec-v2.yaml
kperry-godaddy: Handled here too — both the `422` and `500` now reference the new `Problem` schema. Thanks for flagging both!

internal/adapter/store/sqlitefinder/search.go
kperry-godaddy: Really fair point — thank you, the comment overstated what was happening. Since the finder deliberately carries feed timestamps verbatim (the `ProjectedEntry` doc treats `createdAt` as the authoritative event time), I went with the reword rather than normalizing. It now states the real dependency: the lexical `expires_at` comparison holds because the RA emits canonical UTC RFC 3339 (trailing `Z`, second precision), not because anything normalizes downstream. Thanks for catching the inaccuracy.

internal/config/config.go
```diff
@@ -272,19 +272,32 @@ type Log struct {
...
 
+// EventsFeed configures the public agent-events feed
+// (GET /v1/agents/events) the ANS Finder ingests (RA only).
+type EventsFeed struct {
+	// Retention bounds how far back the feed serves, anchored on each
+	// event's enqueue time. Events older than now-retention age out and
+	// become invisible to the feed. Defaults to 720h (30 days),
+	// matching the production feed's documented retention. A
+	// non-positive value disables the lower bound (serve everything
+	// delivered) — intended for tests only.
```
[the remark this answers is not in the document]
kperry-godaddy: Good catch — the comment promised a config-level disable that Validate actually normalizes away. Reworded in 4b78243 to state the real behavior: a non-positive value is normalized to the 720h default, and the store's no-floor mode is reachable only by constructing the feed store directly, which tests do.

internal/adapter/store/sqlite/outbox.go
```diff
@@ -118,11 +118,19 @@ func (s *OutboxStore) Claim(ctx context.Context, batchSize int) ([]OutboxEvent,
...
+// logID is the value the TL echoed in its ingest response
+// (AppendResult.LogID). It is also echoed on idempotent duplicate
+// retries, so a re-delivered row records the same logId.
+func (s *OutboxStore) MarkSent(ctx context.Context, id int64, logID string) error {
 	_, err := s.db.db.ExecContext(ctx,
-		`UPDATE outbox_events SET sent_at_ms = ? WHERE id = ?`,
-		time.Now().UnixMilli(), id)
+		`UPDATE outbox_events SET sent_at_ms = ?, log_id = ? WHERE id = ?`,
+		time.Now().UnixMilli(), logID, id)
 	return mapSQLErr(err)
```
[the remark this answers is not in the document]
kperry-godaddy: The guard lives at this method's only caller: the outbox worker treats a TL ack with an empty logId as a delivery anomaly — the row is left pending and retried, so an empty value never reaches this UPDATE (pinned by TestWorker_EmptyLogID_KeepsRowPending). Keeping the policy at the single call site and the store a thin persistence layer is deliberate, so no change here.

internal/adapter/store/sqlite/feed.go
```diff
@@ -0,0 +1,166 @@
...
+               COALESCE(r.ans_name, ''), COALESCE(r.agent_host, ''),
+               COALESCE(r.version, ''), COALESCE(r.display_name, ''),
+               COALESCE(r.description, ''),
+               COALESCE(e.endpoints, '')
+        FROM outbox_events o
+        LEFT JOIN agent_registrations r ON r.agent_id = o.agent_id
+        LEFT JOIN agent_endpoints e ON e.agent_id = o.agent_id
+        WHERE o.sent_at_ms IS NOT NULL
+          AND o.log_id IS NOT NULL
+          AND o.created_at_ms >= ?
```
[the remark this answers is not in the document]
kperry-godaddy: An empty-string log_id can't exist in the table: MarkSent is the only writer of the column, and its only caller refuses to mark a row sent when the TL returns an empty logId — the row stays pending for retry (pinned by TestWorker_EmptyLogID_KeepsRowPending). The IS NOT NULL predicate matches that write path, so leaving it as-is.

internal/adapter/store/sqlitefinder/search.go
```diff
@@ -0,0 +1,197 @@
...
+		HasMore:    end < total,
+	}, nil
+}
+
+// matchedRows runs the filtered (and optionally full-text) query and
+// returns every matching candidate. now bounds expiry: a row with a
+// non-empty expires_at at or before now is excluded.
+func (s *Store) matchedRows(ctx context.Context, text string, filter index.Filter, now time.Time) ([]matchedRow, error) {
+	where := []string{"e.lifecycle = 'ACTIVE'", "(e.expires_at = '' OR e.expires_at > ?)"}
+	args := []any{rfc3339(now)}
```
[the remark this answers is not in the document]
kperry-godaddy: This is a documented dependency rather than an oversight: the finder carries feed timestamps verbatim by design, and the lexical compare is sound because the producing RA emits canonical UTC RFC 3339 (trailing Z, second precision) — the rfc3339 helper comment states exactly that, and the non-Z fixture pins verbatim passthrough, not a supported comparison input. datetime() on the column would also bypass the expires_at index on an anonymous endpoint and silently drop rows it can't parse. If the finder ever ingests feeds beyond the RA contract, normalizing at the projection boundary is the fix we'd reach for.

internal/adapter/store/sqlitefinder/explore.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+// matchedSetWhere builds the shared "Active, unexpired, text+filter
+// matched" predicate explore facets aggregate over. Unlike search it
+// expresses the FTS constraint as an EXISTS subquery rather than a join,
+// so a facet that groups by a side-table value does not multiply rows by
+// the FTS join.
+func (s *Store) matchedSetWhere(text string, filter index.Filter, now time.Time) (string, []any, error) {
+	where := []string{"e.lifecycle = 'ACTIVE'", "(e.expires_at = '' OR e.expires_at > ?)"}
+	args := []any{rfc3339(now)}
```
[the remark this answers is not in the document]
kperry-godaddy: Same disposition as the search.go thread: verbatim timestamp carry is deliberate, the producing RA emits canonical UTC Z, and datetime() would trade an index scan for a full scan plus silent NULL-drops on an anonymous endpoint. No change here.

internal/adapter/store/sqlite/sqlite.go
```diff
@@ -136,6 +136,23 @@ func (d *DB) migrate(ctx context.Context) error {
...
+	// here (cheap: the table set is empty or small at migration time)
+	// makes sqlite_stat1 exist, after which the planner uses the partial
+	// `idx_outbox_feed (created_at_ms, id)` index to SEARCH the
+	// retention floor instead. The choice persists as the table grows;
+	// stat1 presence — not its exact row counts — is what flips the
+	// planner off the rowid default.
+	if _, err := d.db.ExecContext(ctx, `ANALYZE`); err != nil {
+		return fmt.Errorf("sqlite: analyze: %w", err)
+	}
 	return nil
```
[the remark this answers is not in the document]
kperry-godaddy: Unconditional is intentional: what matters is that sqlite_stat1 exists — its presence is what flips the planner onto the partial feed index (the comment above walks through it) — and a startup re-run just refreshes statistics during a rare single-writer moment before the server takes traffic. Conditioning on stat1 presence would save little and add a branch for it.

jhateley-godaddy (dismissed):

internal/config/config.go
```diff
@@ -457,6 +484,9 @@ func (c *RAConfig) Validate() error {
 	if c.TLClient.Timeout <= 0 {
 		c.TLClient.Timeout = 10 * time.Second
 	}
+	if c.EventsFeed.Retention <= 0 {
```
cyen-godaddy, line 487: The /v1/agents/events returns events up to 30 days ago, is it possible to return events from the very beginning?

jhateley-godaddy (approved):

internal/config/config.go
kperry-godaddy: I think we can do this as a fast follow if needed.  If users are spinning up the RI though, wouldn't they need to keep it around for 30 days?

PR #47, opened 2026-06-16
feat(ra): AI Catalog generation, FQDN exclusivity, and seal-before-success activation

csnitker-godaddy:
## Summary

Adds producer-side **AI Catalog** artifact generation to the RA, plus the two invariants it depends on. Every byte is derived from the registration aggregate — nothing crawled, nothing fetched — and **the registration request and 202 response are untouched**.

### 1. AI Catalog generation (IMPL slices 1–2)
- **`internal/catalog`** (pure, no I/O): the per-agent `CatalogEntry` and the host-complete `ai-catalog.json` document. Eligibility gates (versioned, ACTIVE, an A2A/MCP endpoint with a policy-passing `metaDataUrl`), single vs nested entries, leftmost-DNS-label URN, `Cc`/`Cf` text sanitization, and an emitted-URL policy (absolute `https`, `host == agentHost`, no userinfo/query/fragment).
- **Routes** (owner-scoped via `ReadOwnership`):
  - `GET /v2/ans/agents/{agentId}/catalog-entry` — the bare entry (`application/json`).
  - `GET /v2/ans/agents/{agentId}/ai-catalog` — the **host-complete document** (`application/ai-catalog+json`) with a strong ETag + `If-None-Match`/304. This is the literal `ai-catalog.json` an Agent-Host Provider republishes at `https://{agentHost}/.well-known/ai-catalog.json`.
- `spec/api-spec-v2.yaml` and the embedded docsui copy updated.

### 2. FQDN exclusivity (one-host-one-owner)
Once a registration is live (ACTIVE/DEPRECATED) on an FQDN, a different owner may not register or activate on it — checked at register, verify-acme, and verify-dns. On activation the winner cancels losing pending registrations (no TL event, ANS-1 §4.4).

### 3. Seal-before-success activation (ANS-1 §12.3)
`verify-dns` seals the single terminal `AGENT_REGISTERED` event **inline** and reports ACTIVE only after the TL acknowledges. A failing or unconfigured sealer **fails closed**: `503 TL_UNAVAILABLE`, the agent stays `PENDING_DNS`, nothing is committed, no outbox row. Revocation still rides the outbox. This is what guarantees a catalogued agent's SCITT-receipt and badge links point at TL records that actually exist.

## The `ai-catalog.json` (the deliverable)

`scripts/demo/ai-catalog.sh` builds one host with a mixed population and saves the validated document to `data/demo/ai-catalog.json`:

| agent | status | in document |
|---|---|---|
| v1.0.0 MCP + metaDataUrl | ACTIVE, eligible | ✅ |
| v2.0.0 A2A + metaDataUrl | ACTIVE, eligible | ✅ |
| v3.0.0 HTTP-API (no card) | ACTIVE, ineligible | ❌ |
| v4.0.0 MCP | PENDING | ❌ |
| eligible agent on a different host | ACTIVE | ❌ |

It asserts host-completeness, the host object, sorted-by-version, per-card `mediaType`, the agent-scoped alias (byte-identical via any agentId), a deterministic ETag, and 304.

## Quality gates
- `make check` green at **90.2%** coverage (`internal/catalog` at 100%); `go test -race` clean on the affected packages.
- Demos pass end-to-end: `run-lifecycle.sh` (V2), `run-lifecycle-v1.sh` (V1), `catalog.sh` (per-agent entry), `ai-catalog.sh` (host document).
- Pre-push **adversarial review** across five dimensions (shape/contract, correctness, security, quality-bar, tests) with each finding independently verified; confirmed findings addressed in this PR (mostly comment/test accuracy), refuted findings documented.

## Known limitations / follow-ups
- **FQDN exclusivity is best-effort under concurrency.** Two different owners each holding a `PENDING_DNS` registration on the same FQDN can race `verify-dns` and both reach ACTIVE (the check runs before the inline seal and outside the activation tx). An in-tx re-check would trade this for a sealed-but-not-activated orphan under seal-before-success, so the proper fix is a **pre-seal host claim** (mirroring the identity lane's nonce claim) — deferred. Documented in `preflightRegistrationConflicts`. The owner-scoped catalog read contains the blast radius.
- **Public discovery route not built.** IMPL §6 models the catalog routes as *public/unauthenticated*; per an earlier maintainer call these are **owner-scoped behind auth**. The genuinely public host-discovery route `GET /v2/ans/catalog?agentHost=` (IMPL slice 3) is unbuilt — that's the surface an AHP fetches without owner credentials.
- Catalog `updatedAt` uses registration/last-renewal time as a documented proxy for the spec's "activation/latest seal timestamp" (the RA persists no activation timestamp; the field is optional).

### DIFF
--- a/internal/catalog/generate.go
+++ b/internal/catalog/generate.go
@@ -1,0 +1,308 @@
+package catalog
+
+import (
+	"strings"
+	"time"
+
+	"github.com/godaddy/ans/internal/domain"
+)
+
+// Options configures CatalogEntry generation.
+type Options struct {
+	// TLPublicBaseURL is the externally-reachable Transparency Log base
+	// URL (e.g. "https://tl.example.org"). When empty — TL not
+	// configured — the entry omits the badge URL and the trust
+	// manifest's attestations rather than emit a malformed URI (§5.2);
+	// the manifest still carries its identity. In production this is
+	// required RA config, so attestations are always present.
+	TLPublicBaseURL string
+	// AllowInsecureURLs permits http metaDataUrl values to pass the URL
+	// policy. A dev-only override (§3.8); false in production, where
+	// only https is emitted.
+	AllowInsecureURLs bool
+}
+
+// NotEligibleError reports that a registration produces no CatalogEntry
+// (IMPL §3.6). The handler maps it to 422 NOT_CATALOG_ELIGIBLE; the
+// registration response uses Reason as the machine-readable omission code.
+type NotEligibleError struct {
+	// Reason is a machine-readable code: ReasonNoVersion,
+	// ReasonNotActive, or ReasonNoEligibleEndpoint.
+	Reason string
+	msg    string
+}
+
+// Machine-readable NotEligibleError reasons.
+const (
+	// ReasonNoVersion: the registration is versionless (§3.6 Gate 1).
+	ReasonNoVersion = "NO_VERSION"
+	// ReasonNotActive: the agent is not ACTIVE. A catalog entry carries an
+	// ANS-Registration attestation (its SCITT receipt) and a TL badge URL,
+	// and those Transparency-Log records only exist once the agent has been
+	// sealed at activation. A pre-activation registration has no TL entry
+	// to link to, so no catalog entry is produced until the agent is live.
+	ReasonNotActive = "AGENT_NOT_ACTIVE"
+	// ReasonNoEligibleEndpoint: no A2A/MCP endpoint carries a
+	// policy-passing metaDataUrl (§3.6 Gate 2).
+	ReasonNoEligibleEndpoint = "NO_ELIGIBLE_ENDPOINT"
+)
+
+func (e *NotEligibleError) Error() string { return e.msg }
+
+// BuildEntry serializes a registration aggregate into its CatalogEntry
+// (IMPL §3). Every field is derived from the aggregate; nothing is fetched.
+// It returns a *NotEligibleError when the registration produces no entry:
+// versionless (§3.6 Gate 1), not ACTIVE (no Transparency-Log entry to link
+// to yet), or no A2A/MCP endpoint with a policy-passing metaDataUrl (§3.6
+// Gate 2). reg.Endpoints must be populated by the caller.
+//
+// The ACTIVE gate is load-bearing, not cosmetic: a catalog entry's trust
+// manifest points at the agent's SCITT receipt and its metadata carries the
+// TL badge URL, and both of those Transparency-Log records are created only
+// when the agent is sealed at activation (verify-dns). Producing an entry
+// for a pre-activation registration would advertise links to TL records
+// that do not exist — so no entry is generated until the agent is live.
+//
+// Endpoint eligibility is evaluated per endpoint (§3.6 pseudocode): HTTP-API
+// and unknown protocols produce no entry; an endpoint with no metaDataUrl
+// produces no entry (no well-known fallback — ANS only points at metadata
+// the registrant actually declared); a metaDataUrl that fails the URL
+// policy (§3.8) is skipped. One eligible endpoint yields a plain top-level
+// entry; two or more yield a nested outer entry (§3.5) whose Data holds a
+// per-protocol child for each, so one agent can expose A2A and MCP without
+// two top-level entries colliding on (identifier, version).
+func BuildEntry(reg *domain.AgentRegistration, opts Options) (Entry, error) {
+	if reg == nil {
+		return Entry{}, &NotEligibleError{
+			Reason: ReasonNoVersion,
+			msg:    "registration is nil",
+		}
+	}
+
+	// Gate 1: versioned only. Every catalog-eligible registration
+	// carries a version (the ANSName is ans://{version}.{agentHost}),
+	// which also keeps (identifier, version) unique by construction
+	// (§3.4). Structurally always true in this RA — there is no
+	// versionless registration path — but enforced honestly for spec
+	// fidelity and any future base-only profile.
+	if reg.AnsName.IsZero() || reg.AnsName.Version().IsZero() {
+		return Entry{}, &NotEligibleError{
+			Reason: ReasonNoVersion,
+			msg:    "registration has no version; only versioned registrations are catalog-eligible",
+		}
+	}
+
+	// Gate: the agent must be ACTIVE — sealed in the Transparency Log — so
+	// the entry's SCITT-receipt attestation and badge URL reference TL
+	// records that actually exist. A pending/failed registration has no TL
+	// entry, so it is not catalogued until it goes live.
+	if reg.Status != domain.StatusActive {
+		return Entry{}, &NotEligibleError{
+			Reason: ReasonNotActive,
+			msg:    "agent is not ACTIVE; a catalog entry is published only after the agent is sealed in the Transparency Log",
+		}
+	}
+
+	agentHost := reg.AnsName.AgentHost()
+	urn := sanitizeText(buildURN(agentHost, deriveLabel(agentHost)))
+
+	// Gate 2: collect endpoints eligible for an entry.
+	eligible := collectEligibleEndpoints(reg.Endpoints, agentHost, opts.AllowInsecureURLs)
+	if len(eligible) == 0 {
+		return Entry{}, &NotEligibleError{
+			Reason: ReasonNoEligibleEndpoint,
+			msg:    "registration has no A2A or MCP endpoint with a metadata URL eligible for cataloging",
+		}
+	}
+
+	entry := Entry{
+		Identifier:  urn,
+		DisplayName: sanitizeText(reg.Details.DisplayName),
+		Description: sanitizeText(reg.Details.Description),
+		Version:     sanitizeText(reg.AnsName.Version().String()),
+		// UpdatedAt: registration (or last-renewal) time is a deliberate
+		// proxy for IMPL §3.1's "activation / latest seal timestamp" — the
+		// RA persists no activation timestamp (Activate discards `now` into
+		// the event only), and updatedAt is optional, so this is a fidelity
+		// proxy, not the sealed go-live time.
+		UpdatedAt:     rfc3339(reg.Details.EffectiveTimestamp()),
+		Publisher:     buildPublisher(agentHost),
+		Metadata:      buildMetadata(reg, opts.TLPublicBaseURL),
+		TrustManifest: buildTrustManifest(urn, reg.AgentID, opts.TLPublicBaseURL),
+	}
+
+	if len(eligible) == 1 {
+		// Single includable endpoint → plain top-level entry (§3.5).
+		entry.MediaType = eligible[0].mediaType
+		entry.URL = eligible[0].url
+		entry.Tags = sanitizeTags(eligible[0].tags)
+		return entry, nil
+	}
+
+	// Multiple includable endpoints → nested outer entry (§3.5). The
+	// outer entry legally uses Data (an ANS-generated nested catalog,
+	// not raw card bytes) and carries the trust manifest; children stay
+	// lean.
+	entry.MediaType = MediaTypeCatalog
+	children := make([]Entry, 0, len(eligible))
+	for _, e := range eligible {
+		children = append(children, Entry{
+			Identifier:  urn + ":" + strings.ToLower(string(e.protocol)),
+			DisplayName: sanitizeText(reg.Details.DisplayName) + " (" + string(e.protocol) + ")",
+			MediaType:   e.mediaType,
+			URL:         e.url,
+		})
+	}
+	entry.Data = &NestedCatalog{SpecVersion: SpecVersion, Entries: children}
+	return entry, nil
+}
+
+// eligibleEndpoint is one endpoint that passed the §3.6 gate.
+type eligibleEndpoint struct {
+	protocol  domain.Protocol
+	mediaType string
+	url       string
+	tags      []string
+}
+
+// collectEligibleEndpoints applies the §3.6 per-endpoint gate, preserving
+// registration order: keep A2A/MCP endpoints whose metaDataUrl is present
+// and passes the URL policy; skip everything else.
+func collectEligibleEndpoints(endpoints []domain.AgentEndpoint, agentHost string, allowInsecure bool) []eligibleEndpoint {
+	out := make([]eligibleEndpoint, 0, len(endpoints))
+	for _, e := range endpoints {
+		mt, ok := protocolMediaType(e.Protocol)
+		if !ok {
+			continue // HTTP-API / unknown → no entry
+		}
+		if e.MetadataURL == "" {
+			continue // no declared metadata → no entry (no fallback)
+		}
+		safeURL, ok := validateEmittedURL(e.MetadataURL, agentHost, allowInsecure)
+		if !ok {
+			continue // fails URL policy → skip endpoint
+		}
+		out = append(out, eligibleEndpoint{
+			protocol:  e.Protocol,
+			mediaType: mt,
+			url:       safeURL,
+			tags:      collectTags(e.Functions),
+		})
+	}
+	return out
+}
+
+// protocolMediaType maps a catalog-eligible protocol to its artifact media
+// type. HTTP-API and any unknown protocol return false — they produce no
+// catalog artifact type (§3.6). The protocol enum is A2A/MCP/HTTP-API;
+// there is no PAYMENT.
+func protocolMediaType(p domain.Protocol) (string, bool) {
+	switch p {
+	case domain.ProtocolA2A:
+		return mediaTypeA2A, true
+	case domain.ProtocolMCP:
+		return mediaTypeMCP, true
+	default:
+		return "", false
+	}
+}
+
+// deriveLabel returns the lineage label for the URN: the leftmost DNS label
+// of agentHost (e.g. ai-agent.acmecorp.com → "ai-agent"). agentHost is
+// RFC-1123 validated upstream (at least two non-empty labels), so the
+// result is always non-empty. The label is stable across versions —
+// agentHost is immutable (ANS-1 §5), and one agentHost is one logical
+// agent (§3.4) — and is keyed on neither agentId (rotates per version) nor
+// displayName (mutable), satisfying §3.3 from data the RA already holds.
+func deriveLabel(agentHost string) string {
+	host := strings.ToLower(strings.TrimSpace(agentHost))
+	if i := strings.IndexByte(host, '.'); i > 0 {
+		return host[:i]
+	}
+	return host
+}
+
+// buildURN composes the domain-anchored lineage handle (§3.3). The host
+// segment is the registration's own agentHost, so a registry reads the
+// publisher domain directly from the identifier.
+func buildURN(agentHost, label string) string {
+	return "urn:ai:" + agentHost + ":agents:" + label
+}
+
+// buildPublisher builds the Publisher block (§3.7). ANS has no verified
+// org name, so identifier and displayName are both the agentHost (honest,
+// no unverified claim) and the identity is DNS-anchored.
+func buildPublisher(agentHost string) *Publisher {
+	host := sanitizeText(agentHost)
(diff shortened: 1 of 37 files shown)


liady:
One small note - ARD [spec v0.9](https://agenticresourcediscovery.org/spec/#42-catalog-entry-object) specifies `urn:air:` as the identifier prefix instead of`urn:ai:`.

internal/ra/service/registration.go
```diff
@@ -482,6 +545,236 @@ func (s *RegistrationService) resolveServerCertInput(
...
+// cancellations carry no TL event (ANS-1 §4.4). Caller runs this inside
+// the activation transaction so winner-activates and losers-cancel commit
+// atomically.
+func (s *RegistrationService) cancelConflictingPendings(ctx context.Context, agentHost, winnerOwnerID string) error {
+	regs, err := s.agents.FindAllByAgentHost(ctx, agentHost)
+	if err != nil {
+		return err
+	}
+	for _, r := range regs {
+		if r.OwnerID == winnerOwnerID || !r.Status.IsPending() {
```
kperry-godaddy, line 664: The doc comment on `preflightRegistrationConflicts` is admirably honest about the pending-window race, but I traced two consequences that make the window worth closing (or at least detecting) before this lands:

1. This skip condition treats a rival's *committed ACTIVE* row as ignorable. If owners A and B race verify-dns, both pass the pre-seal gate; when B commits first, A's transaction arrives in this loop holding B's now-ACTIVE row in hand — and skips it instead of aborting. Both end up ACTIVE, each with a permanent `AGENT_REGISTERED` leaf.
2. The SQLite `Save` UPDATE has no status guard (`WHERE id = ?` only, `internal/adapter/store/sqlite/agent.go`), so in the interleaving where B's tx conflict-cancels A mid-seal, A's tx then re-saves its stale in-memory ACTIVE aggregate and silently resurrects the row the cancellation just revoked.

Once both are ACTIVE, `ensureHostNotTakenByOther` returns 409 for *each* owner against the other on every subsequent call — the host is wedged for both until manual DB surgery, and nothing detects the violated invariant.

Since the store runs a single writer, an in-tx re-check is deterministic and cheap: treat any other-owner row with `holdsHostExclusivity(r.Status)` in this loop as `errHostTaken` (aborting the activation), and re-read this registration's own row inside the tx, requiring it still be PENDING_DNS before `Save` — the same conditional-commit shape the identity lane already uses. The loser's already-sealed leaf then falls into the same benign-residue class the comments already accept for commit failures, and the pre-seal host claim can land as the durable fix on its own schedule.

Related: the spec's new 409 descriptions state one-host-one-owner unconditionally ("the FQDN belongs to that owner until no live registration remains"), while this comment correctly says best-effort. Worth aligning the two whichever way we land.

internal/ra/service/registration.go
```diff
@@ -482,6 +545,236 @@ func (s *RegistrationService) resolveServerCertInput(
...
+			continue
+		}
+		if err := r.CancelForHostConflict(); err != nil {
```
kperry-godaddy, line 726: By the time a registration reaches PENDING_DNS it has already had its identity CSR signed at verify-acme, so a loser cancelled here holds a VALID identity certificate. `cancelPending` — the sibling path for the same pending states — revokes those certs at the CA and flips the cert rows in the same operation; this path skips both. And the state is unreachable afterward: `Revoke` on a now-REVOKED registration takes the idempotent early return without touching certs. Net effect: a registration we've decided lost its host keeps a live credential asserting an ANS name on that host for up to its 90-day term, the cert store reports VALID certificates for a REVOKED agent, and no API path can fix it.

Suggest mirroring `cancelPending`'s structure: collect the losing registrations before the activation tx opens (the pre-seal host read already fetches them), run `revokeIdentityCertsAtCA` outside the tx (idempotent by port contract — the same placement the revoke flow uses), and flip the VALID cert rows to REVOKED inside the tx alongside the cancellation. Making the idempotent-revoke branch sweep any still-VALID identity certs would also make this state self-healing if one ever slips through.

internal/ra/handler/catalog.go
```diff
@@ -0,0 +1,150 @@
...
+// CatalogEntry is the registrant's own view of how their agent will appear
+// in a catalog. The public, crawlable surface is the population export
+// (a later slice) and the well-known per-host document an AHP publishes —
+// not these agent-keyed routes.
+type CatalogHandler struct {
+	svc *service.RegistrationService
+}
+
+// NewCatalogHandler constructs a CatalogHandler.
+func NewCatalogHandler(svc *service.RegistrationService) *CatalogHandler {
```
kperry-godaddy: Nice, compact handler — the marshal-before-headers ordering in `writeCatalogDocument` and the RFC 9110 weak-comparison If-None-Match handling are both exactly right. One structural item before merge: this constructor takes no logger, and the error paths below all go through the package-level `WriteError`, which `errors.go` documents as the non-logging entry point for callers *outside* the package. Every sibling handler embeds `responder` and routes through `h.writeError` precisely because the RA has no request-log middleware and the service layer doesn't log reads — that seam is the only place a non-domain 500's cause gets recorded. As written, a store fault on either catalog route returns the sanitized generic 500 with zero server-side trace.

Mechanical fix:

```go
type CatalogHandler struct {
	responder
	svc *service.RegistrationService
}

func NewCatalogHandler(svc *service.RegistrationService, logger zerolog.Logger) *CatalogHandler {
	return &CatalogHandler{responder: newResponder(logger), svc: svc}
}
```

…then `h.writeError` at the six `WriteError` call sites (threading the responder into `writeCatalogDocument`), and pass the logger at the `main.go` construction site. The 422 `NOT_CATALOG_ELIGIBLE` path is an expected 4xx and can stay as-is.

internal/ra/service/registration.go
```diff
@@ -482,6 +545,236 @@ func (s *RegistrationService) resolveServerCertInput(
...
+	}
+
+	producerSig, err := s.signCanonical(ctx, innerCanonical, now)
+	if err != nil {
+		return nil, fmt.Errorf("sign agent event: %w", err)
+	}
+
+	sealCtx, cancel := context.WithTimeout(ctx, s.sealTimeout)
+	defer cancel()
+	logID, err := s.agentSealer.SealAgentEvent(sealCtx, schemaVer, innerCanonical, producerSig)
```
kperry-godaddy, line 834: The fail-closed shape here is exactly right — nil sealer and seal failure both leave the agent PENDING_DNS with nothing committed, and running the seal outside the write lock was the correct call. What's missing is the operator signal: this call currently succeeds or fails silently. The identity lane this mirrors logs both outcomes (`identity.go:938-954` — INFO with identifiers on success, WARN/ERROR with the cause on failure), and CLAUDE.md treats "agent activated / event appended" as required INFO events. Three concrete asks:

1. INFO on seal success with `agentId`, `fqdn`, `schemaVersion`, `logId`, and duration — logged before the commit, which also makes any orphaned leaf from a commit failure traceable back to us.
2. WARN (transient) / ERROR (permanent) with `.Err(err)` on seal failure. Today the only line anywhere is the handler's generic 500/503 log, and `SealAgentEvent`'s transient branch maps through `NewUnavailableError(code, message)`, which drops the underlying cause — so during a TL incident we can't distinguish timeout from connection-refused from 429 from the RA side. Worth carrying the cause through, or logging it at the adapter boundary before mapping.
3. One INFO per registration cancelled in `cancelConflictingPendings` (`loserAgentId`, `winnerAgentId`, `fqdn`) — it's a state transition on another owner's resource, and it's the first thing support will need when that owner asks why their pending registration turned REVOKED.

Giving `RegistrationService` the standard `WithLogger` option (component-tagged, `zerolog.Nop()` default) would cover all three per repo convention.

internal/catalog/generate.go
```diff
@@ -0,0 +1,334 @@
...
+	// Multiple includable endpoints → nested outer entry (§3.5). The
+	// outer entry legally uses Data (an ANS-generated nested catalog,
+	// not raw card bytes) and carries the trust manifest; children stay
+	// lean.
+	entry.MediaType = MediaTypeCatalog
+	children := make([]Entry, 0, len(eligible))
+	for _, e := range eligible {
+		children = append(children, Entry{
+			Identifier:  urn + ":" + strings.ToLower(string(e.protocol)),
+			DisplayName: sanitizeText(reg.Details.DisplayName) + " (" + string(e.protocol) + ")",
```
kperry-godaddy: Small contract edge on the nested children: `displayName` can be a full 64 characters at registration (the domain cap), and this suffix adds up to 6 more — while the children are `$ref CatalogEntry` in the spec, whose `displayName` declares `maxLength: 64`. So a legal registration with a 59+ character name and two eligible protocols produces a 200 body that fails validation against our own published schema, and a schema-validating consumer will reject the whole document.

Two cheap outs: drop the suffix — the child is already fully discriminated by its `:a2a`/`:mcp` identifier segment and its `mediaType`, so the suffix carries no information — or keep it and remove `maxLength` from the response schema (loosening a response constraint is a compatible change; the registration request is where the cap is actually enforced). I'd lean toward dropping the suffix so parent and child stay byte-identical on the name.

internal/catalog/urlpolicy.go
```diff
@@ -0,0 +1,62 @@
...
+//     consumer at a metadata document on a host the agent does not own.
+//
+// The Transparency-Log badge and receipt URLs are RA-controlled and built
+// from already-validated config (PublicBaseURL), not registrant input, so
+// they do not pass through this agentHost-bound check.
+func validateEmittedURL(raw, agentHost string, allowInsecure bool) (string, bool) {
+	if raw == "" || agentHost == "" {
+		return "", false
+	}
+	u, err := url.Parse(raw)
```
kperry-godaddy, line 42: One check went missing in the port from the finder: `internal/finder/project/sanitize.go`'s `validateEmittedURL` rejects any Cc/Cf rune anywhere in the raw URL before parsing — with a good comment explaining why URLs must be rejected rather than stripped — and this re-implementation dropped that loop. `url.Parse` only rejects ASCII controls, so a bidi override or zero-width rune in the *path* parses cleanly, passes the host pin, and ships verbatim into a document that gets rendered in UIs and LLM contexts. It's unreachable through today's registration path (`validateMetadataURL` enforces printable ASCII), so this is defense-in-depth — but it contradicts `sanitize.go`'s "every emitted string passes this one chokepoint" doc, and rows persisted before that registration-time check existed wouldn't be covered. One line closes it:

```go
if strings.ContainsFunc(raw, isControlOrFormat) {
	return "", false
}
```

Bigger picture: this policy, `sanitizeText`, and `deriveLabel`/`buildURN` are deliberate byte-for-byte mirrors of the finder's copies, held in lockstep by comments — and the copies have already diverged twice inside this PR's lifetime (the `urn:ai` → `urn:air` fix touched only this package; this missing rune check is the other instance). Since "search results and the published catalog hand consumers one identifier" is a stated invariant, I'd love to see the shared trio extracted into one package both sides import — or at minimum a cross-package test pinning `deriveLabel ≡ labelize` and the URN format — before the public host route builds on it.

spec/api-spec-v2.yaml
```diff
@@ -375,6 +381,158 @@ paths:
...
+        '422':
+          description: |
+            Registration is not catalog-eligible: not ACTIVE (no
+            Transparency-Log entry to link to yet), versionless, or no
+            A2A/MCP endpoint with a policy-passing metaDataUrl. `code` is
+            `NOT_CATALOG_ELIGIBLE`; `detail` names the specific gate.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
```
kperry-godaddy: The prose here is telling the truth and the schema isn't: `detail` *is* what the server sends — every failure path emits RFC 7807 `application/problem+json` with `{type, title, status, detail, code}` (`writeProblem` in `internal/ra/handler/errors.go`) — but the declared content is `application/json` with `ErrorResponse` (`{status: "ERROR", code, message, details}`), a shape no handler produces. A client generated from this spec can't parse a single error the server returns: wrong media type, wrong field names, wrong `status` type. `ErrorResponse` predates this PR, but this diff adds roughly fifteen new declarations against it (both catalog routes' 4xx/5xx, the register/verify 409s, the verify-dns 503).

For this PR's scope: add a `Problem` schema — `{type: string, title: string, status: integer, detail: string, code: string}` — and point the new routes' error responses at it as `application/problem+json`. Migrating the pre-existing declarations can be its own follow-up; I'd just rather not extend the drift to fifteen more sites, on routes where we're explicitly documenting `code`-based client handling.

kperry-godaddy (commented):
Connor — this is a strong PR. The seal-before-success mechanics are genuinely well built (fail-closed nil-sealer handling, the seal running outside the write lock, byte-exact payload persisted onto the pre-delivered feed row), the host document is correctly owner-scoped at both the handler and service layers, the byte-pinned golden test against the spec is exactly how we keep wire shapes honest, and the test-enforced docsui sync eliminated a whole class of review work. The comments in the tricky spots are candid about their trade-offs, which made this a pleasure to review.

I've left seven inline comments on things I'd like us to resolve before merge: two correctness edges on the new exclusivity invariant (the race window has consequences beyond the documented one, including a certificate-lifecycle gap), two observability gaps on the new activation lane, and three wire-contract items. Every one of them looks like a contained, mechanical fix — no redesign anywhere. Happy to pair on any of them.

internal/ra/service/registration.go
csnitker-godaddy: Fixed in 82cac02. Replaced the blind post-seal `Save` with `commitActivation`, which re-reads this row and the host's rows **inside** the activation tx (single writer -> deterministic): it aborts with `AGENT_HOST_TAKEN` if a rival committed ACTIVE/DEPRECATED during the seal round trip, and requires this registration's own row to still be `PENDING_DNS` before `Save`, so a rival's mid-seal conflict-cancel can't be resurrected from the stale in-memory ACTIVE aggregate. The loser's already-sealed leaf falls into the benign-residue class you noted. `activation_race_test.go` pins both interleavings via a seal-round-trip hook. The `preflightRegistrationConflicts` comment and the spec's 409 text now describe activation-time enforcement as authoritative; a pre-seal host claim is left as an optional optimization rather than the correctness fix.

internal/ra/service/registration.go
csnitker-godaddy: Fixed in 82cac02. `commitActivation` now mirrors `cancelPending`: `revokeConflictLoserCertsAtCA` revokes each loser's identity certs at the CA before the tx (idempotent per the port contract), and the VALID cert rows flip to REVOKED inside the tx alongside the cancellation. `Revoke`'s idempotent already-REVOKED branch now calls `sweepRevokedAgentCerts`, so a cert that slips through the cancellation window self-heals on a later revoke instead of being unreachable. Covered by `TestVerifyDNS_ConflictCancelRevokesLoserCerts` and `TestRevoke_IdempotentSweepsLingeringCerts`.

internal/ra/handler/catalog.go
csnitker-godaddy: Done in 82cac02. `CatalogHandler` now embeds `responder` and takes a `zerolog.Logger`; the six `WriteError` sites are `h.writeError` (threaded into `writeCatalogDocument`), and `main.go` passes the logger at construction. The 422 `NOT_CATALOG_ELIGIBLE` path stays a bare 4xx as you called out.

internal/ra/service/registration.go
csnitker-godaddy: Done in 82cac02. `RegistrationService` gained `WithLogger` (component-tagged, `zerolog.Nop()` default): (1) INFO on seal success with `agentId`/`fqdn`/`schemaVersion`/`logId`/`sealDuration`, logged before the commit so an orphaned leaf is traceable; (2) WARN (transient) / ERROR (permanent) with `.Err(err)` on seal failure, and the raw transport cause is now logged at the tlclient boundary (`logSealFailure`) before `NewUnavailableError` strips it; (3) INFO per conflict-cancelled loser (`loserAgentId`/`winnerAgentId`/`fqdn`) in `commitActivation`.

internal/catalog/generate.go
csnitker-godaddy: Fixed in 82cac02 — dropped the suffix, as you preferred. Nested children reuse the parent `displayName` verbatim; they stay fully discriminated by their `:a2a`/`:mcp` identifier segment and `mediaType`, and a 64-char registered name can no longer overflow the published `CatalogEntry` cap and self-invalidate the document.

internal/catalog/urlpolicy.go
csnitker-godaddy: Both done in 82cac02. `validateEmittedURL` now rejects any Cc/Cf rune in the raw URL before parsing (`strings.ContainsFunc(raw, isControlOrFormat)`), matching the finder. And the URN derivation is extracted: `internal/ard` (`MintURN`/`Labelize`) is imported by both `internal/finder/project` and `internal/catalog`, so the one-identifier invariant is structural — pinned by `internal/ard` tests plus a catalog seam test. Scope note: I folded in the URN mint (the piece the invariant hangs on) and fixed the URL-policy divergence, but `sanitizeText` still lives as a copy in each package; happy to pull it into `internal/ard` too in a follow-up if you'd like the whole trio shared.

spec/api-spec-v2.yaml
csnitker-godaddy: Done in 82cac02. Added a `Problem` schema (`{type, title, status, detail, code}`) and pointed this PR's new error responses at it as `application/problem+json`: both catalog routes' 4xx/5xx, the register and verify-acme 409s, and the verify-dns 409/503. The pre-existing `ErrorResponse` declarations are left for the separate follow-up you suggested (a note on the new `Problem` schema records that).

internal/ra/handler/lifecycle_test.go
```diff
@@ -809,9 +862,146 @@ func TestRevoke_NotOwnedReturns403(t *testing.T) {
...
 	svc    *service.RegistrationService // exposed for direct-handler tests
 }
 
+// ----- FQDN exclusivity (one-host-one-owner) -----
+//
+// Once a registration goes live (ACTIVE/DEPRECATED) on an FQDN, that FQDN
+// belongs to its owner alone: a different owner may neither register nor
+// activate on it until no live registration remains. Enforced at
+// registration (entry gate) and re-checked atomically at activation
+// (verify-dns) to close the pending-window race.
```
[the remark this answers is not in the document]
csnitker-godaddy: Addressed — the test comment no longer asserts an atomic guarantee. As of 82cac02 the activation exclusivity check IS authoritative at commit time (`commitActivation` re-checks in-tx under the single writer), and the comment now describes that alongside the per-step fast-fail checks.

kperry-godaddy (approved):
Approving — this is excellent follow-through, Connor. Every item landed as a real fix rather than a patch-over:

- `commitActivation` makes host exclusivity authoritative at commit time, and the new race tests pin both interleavings (including the no-wedge retry) via the seal-round-trip hook — really nice test design.
- The loser certificate lifecycle now mirrors `cancelPending` end to end, and the self-healing sweep turns the previously-unreachable state into a repair path.
- `internal/ard` is the right structural answer to the duplication: the one-identifier invariant is now enforced by the import graph instead of by comments.
- The seal/cancellation logging (with the transport cause preserved at the client boundary) and the `Problem` schema both landed exactly where they needed to.

Verified on my side: `make check` green at 90.7% coverage, `go test -race` clean across every touched package, and the docsui copy still byte-identical to `spec/`.

Two tiny non-blocking notes so they're recorded for a follow-up: (1) a rival whose verify-acme commits inside the window between the pre-tx CA pass and the activation transaction gets its cert rows correctly flipped in-store but misses the CA-side revocation — and the sweep can't see that case, since it keys on store-VALID rows. Having `commitActivation` return the certs it flipped and CA-revoking exactly that set post-commit (idempotent) would close it completely. (2) A concurrent duplicate verify-dns for the *same* agent now surfaces the second commit as `AGENT_HOST_TAKEN`, which reads as "another operator" for your own now-ACTIVE agent — cosmetic, and a retry self-corrects via the idempotent branch. Neither needs to hold this up. And yes to your offer on pulling `sanitizeText` into `internal/ard` — that finishes the trio.

Great work. :ship:

PR #68, opened 2026-07-14
Fix/align specs with implementation

### LINKED ISSUE #59: docs: RA/TL OpenAPI and served TL event schemas drift from shipped behavior
Found during the spec-conformance review of agentnameservice/ans-registry#32 ([review](https://github.com/agentnameservice/ans-registry/pull/32#pullrequestreview-4648621233)). Evidence pinned to [`974ca94`](https://github.com/agentnameservice/ans/commit/974ca9470dcf6e239250ea6c43cd44970c03c53d). Line refs to `spec/api-spec-v2.yaml` unless noted; the docsui mirrors (`internal/adapter/docsui/openapi/*.yaml`) need the same edits.

Collection of places where the published OpenAPI/JSON-schema contracts and the shipped behavior disagree. Each is small; batched here so they can be triaged together.

### Contract says one thing, code does another

1. **Register prose: "The request MUST include an identity CSR"** (`api-spec-v2.yaml:181-183`). The schema itself marks `identityCsrPEM` optional ("When omitted, the agent registers without an identity certificate", :1566-1572) and the code enforces exactly-one-of `serverCsrPEM`/`serverCertificatePEM` instead ([`registration.go#L423-L430`](https://github.com/agentnameservice/ans/blob/974ca9470dcf6e239250ea6c43cd44970c03c53d/internal/ra/service/registration.go#L423-L430)). The prose predates identity-optional registration. (The registration schema also lacks the `oneOf` the renewal schema has at :2008-2010.)
2. **Enum casing: responses violate the published enums.** The yaml declares `Protocol: [A2A, MCP, HTTP-API]` (:1473-1475) and hyphenated transports `[STREAMABLE-HTTP, SSE, JSON-RPC, ...]` (:1886-1890), but the RA canonicalizes, stores, and emits the underscore forms `HTTP_API`, `STREAMABLE_HTTP`, `JSON_RPC` ([`protocol.go#L11-L15`](https://github.com/agentnameservice/ans/blob/974ca9470dcf6e239250ea6c43cd44970c03c53d/internal/domain/protocol.go#L11-L15), `#L50-L57`). Parsers accept both on input; output should match the declared enum (or the enum should match the output).
3. **TL yaml still lists the retired event model.** `spec/api-spec-tl-v2.yaml:1000` — `eventType: [AGENT_REGISTRATION, DOMAIN_VALIDATION, AGENT_ACTIVE, AGENT_REVOCATION]`. Shipped enums are `AGENT_REGISTERED / AGENT_RENEWED / AGENT_DEPRECATED / AGENT_REVOKED` ([`event.go#L72-L102`](https://github.com/agentnameservice/ans/blob/974ca9470dcf6e239250ea6c43cd44970c03c53d/internal/tl/event/event.go#L72-L102)).
4. **`V2.json` omits `dnssecVerified`.** The V2 builder seals per-record `dnssecVerified` ([`lifecycle.go#L1060-L1078`](https://github.com/agentnameservice/ans/blob/974ca9470dcf6e239250ea6c43cd44970c03c53d/internal/ra/service/lifecycle.go#L1060-L1078); `event.go:201-213`), but the served `DNSRecord` schema (`internal/tl/service/schemas/V2.json:107-129`) has no such property.
5. **`revocationReasonCode` enum mismatch, unenforced.** The served schemas allow 11 tokens including `UNSPECIFIED`, `REMOVE_FROM_CRL`, `CA_COMPROMISE`, `EXPIRED_CERT` (`V2.json:184-196`), the RA emits only the 7-token set (`status.go:141-149`; `api-spec-v2.yaml:1526-1535`), and TL ingest validates the field not at all (`event.go:263-283` requires only eventType/ansId/ansName/timestamp — any string seals). Worth deciding the canonical set and enforcing it at ingest.

### Constraint drift within the request schema

6. **`AgentFunction`**: yaml caps `id`/`name` at 64 chars and `tags` at 5 (:1895-1912); Go enforces none of those, instead non-blank id/name, unique ids per endpoint, and ≤20 tags of ≤20 chars ([`endpoint.go#L31-L53`](https://github.com/agentnameservice/ans/blob/974ca9470dcf6e239250ea6c43cd44970c03c53d/internal/domain/endpoint.go#L31-L53), `#L103-L116`).
7. **`AgentLifecycleStatusFilter`** enum is `[PENDING_DNS, ACTIVE, DEPRECATED, REVOKED, ALL]` (:1740-1742) — `PENDING_VALIDATION`/`FAILED`/`EXPIRED` are unfilterable per the contract, and the handler doesn't validate the value at all (arbitrary strings pass through).
8. **Enforced in code, absent from the contract**: at most one endpoint per protocol (422 `DUPLICATE_PROTOCOL`, [`endpoint.go#L191-L197`](https://github.com/agentnameservice/ans/blob/974ca9470dcf6e239250ea6c43cd44970c03c53d/internal/domain/endpoint.go#L191-L197)); `metaDataHash` requires `metaDataUrl` (`endpoint.go:88-94`).
9. **Declared in the contract, unenforced in code**: `agentDisplayName` required (:1605-1609 — code checks only the 64-char max and accepts empty, `agent.go:124-129`); revocation `comments` `maxLength: 200` (:1616-1618 — no runtime check at any layer).

Suggest one pass that picks the source of truth per item (mostly: make the yaml match shipped behavior; for 5 and 9, decide whether to enforce).

jhateley-godaddy:
Fixes schema/implementation drift issues reported in https://github.com/agentnameservice/ans/issues/59.

**Validation and Error Handling Improvements:**
- Added required validation for `displayName` in `NewRegistration` and corresponding test coverage for missing and overly long values. [[1]](diffhunk://#diff-e7648404262e30216b5f15acfa4dcc59b58d2a540b9d8bdd0e82953fd1ff2956R121-R123) [[2]](diffhunk://#diff-5fd3d67c42c965af90c23784cf35298944d72370ec7a762622092deb5b9c8d39L88-R94)
- Added length checks for `comments` in revocation requests, returning a 422 error if exceeded, with tests for this case. [[1]](diffhunk://#diff-af0e6df1e96c5af936e71bc86e98043bdd5eb2af852c1f554fcd099d319bd49fR157-R162) [[2]](diffhunk://#diff-318b4a011db12f05fef395ed9994958394e1d225d06a7f7eb173a22d9322ed29R388-R405)
- Added validation for `RevocationReasonCode` in event payloads and associated tests. [[1]](diffhunk://#diff-d428d4f6a2eba95bdefd5e42643769efe408300bebebb1fcd17567b1dda84ccaR283-R285) [[2]](diffhunk://#diff-8a0ac4d5ac78d9d0763b2ffb67706c0b7f11c07fd31420313df6b64be7b8d718R124-R132)
- Limited `AgentFunction` tags to a maximum of 5, and added max length checks (64 chars) for both `ID` and `Name`, with updated validation and tests. [[1]](diffhunk://#diff-ee330b4d32eb37f0692f0a8ecdb3361e618af4d755d7efbf58bdbb002535ab3dL12-R15) [[2]](diffhunk://#diff-ee330b4d32eb37f0692f0a8ecdb3361e618af4d755d7efbf58bdbb002535ab3dR34-R53) [[3]](diffhunk://#diff-9ac30271dcc984253c3d001aa27ee6aa6e6956e5162721bd281f6c9e1898b9f1R17-R31) [[4]](diffhunk://#diff-9ac30271dcc984253c3d001aa27ee6aa6e6956e5162721bd281f6c9e1898b9f1L33-R43)

**Enum and OpenAPI Schema Updates:**
- Normalized enum values for `Protocol` and transport types (e.g., `HTTP-API` → `HTTP_API`, `STREAMABLE-HTTP` → `STREAMABLE_HTTP`) in OpenAPI specs and parsing logic. [[1]](diffhunk://#diff-dc5746e494bc76e1055d546fb66b8ce2973c5a56e9776593712f3943c5540b6eL1475-R1474) [[2]](diffhunk://#diff-dc5746e494bc76e1055d546fb66b8ce2973c5a56e9776593712f3943c5540b6eL1890-R1915) [[3]](diffhunk://#diff-17d322da85dca0e5d9ae31d97b069bf855ebfccc6c2ec13d47a2684950475182R63)
- Updated event type enums for agent lifecycle events to use past-tense and standardized values.

**OpenAPI/Schema Clarifications:**
- Improved documentation and schema for agent registration, including clarifying CSR requirements, endpoint uniqueness, and required fields. [[1]](diffhunk://#diff-dc5746e494bc76e1055d546fb66b8ce2973c5a56e9776593712f3943c5540b6eL177-R182) [[2]](diffhunk://#diff-dc5746e494bc76e1055d546fb66b8ce2973c5a56e9776593712f3943c5540b6eR1561-R1568) [[3]](diffhunk://#diff-dc5746e494bc76e1055d546fb66b8ce2973c5a56e9776593712f3943c5540b6eR1621-R1623)
- Added documentation and schema constraints for `metaDataHash` and its relationship with `metaDataUrl`, and enforced their co-presence. [[1]](diffhunk://#diff-dc5746e494bc76e1055d546fb66b8ce2973c5a56e9776593712f3943c5540b6eR1891-R1894) [[2]](diffhunk://#diff-dc5746e494bc76e1055d546fb66b8ce2973c5a56e9776593712f3943c5540b6eL1890-R1915)

**API Filtering and Error Reporting:**
- Restricted filterable agent statuses in the lifecycle handler to only valid, supported values, returning 422 for invalid or unfilterable statuses, with new tests for these cases. [[1]](diffhunk://#diff-db7914c61325b22e9c06d5769826ad8a83b394802c80b78f68f060fe2a7aac58R65-R72) [[2]](diffhunk://#diff-a8feab5606290789141aacaa422d47809ea63c524441731da2096569679c05eaR123-R142)

**Other Minor Improvements:**
- Added constants for maximum lengths to improve maintainability.


jcrisp-godaddy (commented):
Curious about the inclusion of `EXPIRED_CERT` in the revocation reason enum — it doesn't appear in RFC 5280's CRLReason enumeration (Section 5.3.1), which the V2.json schema cites as its source. Is there a use case, or did it perhaps come along with the original V2.json set?

jcrisp-godaddy (commented):
`REMOVE_FROM_CRL` is in RFC 5280, but it's an un-revocation signal (reinstating a cert from `CERTIFICATE_HOLD`) rather than a reason to revoke. Even if ANS eventually verifies external CRLs, that would be a read operation — the ANS revocation reason would still reflect the underlying cause (e.g. `KEY_COMPROMISE`), not a CRL operational directive. Is there a "reinstate from hold" lifecycle in mind that would give this code meaning, or is it safe to drop?

jcrisp-godaddy (commented):
The original 7-token set appeared to deliberately exclude `UNSPECIFIED`, which has the effect of requiring callers to state a reason when revoking. Including it weakens the audit trail — a revocation without a meaningful reason is harder to act on or investigate later. Was the exclusion intentional, and is there a case where a caller genuinely can't determine a reason?

jhateley-godaddy:
@jcrisp-godaddy I agree with your findings.

`REMOVE_FROM_CRL`
From RFC 5280: 
"The removeFromCRL (8) reasonCode value may only appear in delta CRLs and indicates that a certificate is to be removed from a CRL because either the certificate expired or was removed from hold."
- This isn't a CRL, doesn't make sense here.

`EXPIRED_CERT`
- correct, this does not appear in the RFC 5280.  I do see a use case where the agent is revoked due to cert expiry - perhaps this can be added later if necessary?

`UNSPECIFIED`
- I think the exclusion was intentional for the reasons you mentioned.

In short - I will revert the change to align with the original 7 and remove the RFC 5280 reference.  `SUPERSEDED` is already a divergence from the RFC.

- KEY_COMPROMISE
- CESSATION_OF_OPERATION
- AFFILIATION_CHANGED
- SUPERSEDED
- CERTIFICATE_HOLD
- PRIVILEGE_WITHDRAWN
- AA_COMPROMISE

internal/ra/handler/v1lifecycle.go
```diff
@@ -149,6 +154,12 @@ func (h *V1LifecycleHandler) Revoke(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if len(req.Comments) > maxCommentsLength {
+		WriteError(w, domain.NewValidationError("COMMENTS_TOO_LONG",
+			fmt.Sprintf("comments exceeds %d characters", maxCommentsLength)))
+		return
+	}
```
csnitker-godaddy, line 163: **Major — this guard needs a twin on the V2 lane (or a home in the shared service).**

The declaring spec for this constraint is `spec/api-spec-v2.yaml` (`AgentRevocationRequest` → `comments: maxLength: 200`) — the exact line issue #59 item 9 cites — but `POST /v2/ans/agents/{agentId}/revoke` (`internal/ra/handler/lifecycle.go:375-392`) decodes `comments` with no length check at any layer: the V2 handler validates only `reason`, `service.Revoke` never reads `RevokeInput.Comments`, and `domain.Revoke(reason, now)` takes no comments parameter. Any value up to the route's 64 KiB body cap returns 200 OK, so the same logical request is 422 on V1 but 200 on V2, and item 9's second half stays open on the only lane whose published spec declares the constraint.

Suggested fix: mirror this check in `LifecycleHandler.Revoke` after the `MISSING_REASON` guard (the constant is already in this package), plus a V2 twin of `TestV1Revoke_CommentsTooLong_422`. Alternative worth considering: enforce once in `service.Revoke`, since `RevokeInput` already carries the field for both lanes — that placement makes this class of single-lane miss impossible, the same reasoning that put the `displayName` check in `domain.NewRegistration`.

csnitker-godaddy (commented):
# PR Review — intent-focused pass on #68

**Scope:** `fix/align-specs-with-implementation` → `main`, 14 commits (13 change commits + 1 no-op merge of main), +194/−46 across 20 files. Two directions of spec↔code alignment closing issue #59's nine drift items: spec edits (`spec/api-spec-v2.yaml`, `spec/api-spec-tl-v2.yaml`, docsui mirrors, served TL schemas V0/V1/V2.json) and new runtime validation (domain, RA handlers, TL ingest codecs on both lanes). No new dependencies. CI green on head `1c462fe`.

**Process.** Three passes: (1) three reviewers each answering a slice of ten intent-derived questions produced 9 findings; (2) verification against the checked-out head (call-graph walks for `Validate()` reach, `.Comments` readers, schema serving, spec property definitions) dropped 3, leaving 6; (3) an adversarial cross-examiner (killed 2) and an independent from-scratch reviewer (raised exactly 1, matching the survivors' top item) reconciled to **1 final finding**. Everything else is in the drop list below with the criterion that removed it — several are one-line fixes worth cherry-picking anyway (marked ⚙).

---

## Recommendation

**Merge after fixing the one finding below** — roughly 6 lines of production code plus one test, completing the PR's own stated deliverable. Also worth updating the PR description before merge: it still advertises the RevocationReason enum *expansion* that `a7d3353` reverted (the final diff *removes* those 4 tokens from the served schemas), so the merge commit reference would be misleading as-is.

This is careful work: 8½ of issue #59's nine items are genuinely closed, both docsui mirrors are byte-identical to the spec files (verified by blob hash), the TL ingest guard is symmetric across the V1/V2 lanes with the enum delegated to the single domain source of truth, and the tightened limits provably cannot retro-break agents persisted under the old limits (validation is entry-gated at `NewRegistration`; no rehydration/renewal/verify path re-validates stored data). The one miss: the `comments` length cap landed on the lane whose spec doesn't declare it, and not on the lane whose spec does.

---

## Finding

### 1. V2 revoke lane never enforces the spec-declared `comments` maxLength 200 — **Major**

See the inline comment on `internal/ra/handler/v1lifecycle.go:157-161` for full detail and the suggested fix. Short version: `spec/api-spec-v2.yaml` `AgentRevocationRequest` declares `comments: maxLength: 200` (the exact line issue #59 item 9 cites), but `POST /v2/ans/agents/{agentId}/revoke` accepts any comments value up to the 64 KiB body cap with 200 OK — no layer below the handler checks it either. The new `COMMENTS_TOO_LONG` guard exists only on the V1 lane.

---

## What we looked at and dropped

Documented so these don't get re-raised later. ⚙ = cheap fix, author's discretion.

- ⚙ **Served historical schemas narrowed while the spec still claims they're verbatim production fetches.** V0.json/V1.json `revocationReasonCode` went 11→7 tokens, but `spec/api-spec-tl-v2.yaml:666-669` still says the V0/V1 bodies "are the production schemas fetched from transparency.ans.godaddy.com… for historical-verifier backwards compatibility" — now false, and V0 has no ingest lane in this TL (the file exists purely as a fidelity mirror; its retired eventType tokens were deliberately kept in the same edit). Cheap fix: reword the route description (and consider restoring V0's historical enum).
- ⚙ **Register description prose still says server-cert input is "optional".** The same PR added `oneOf` making exactly-one-of `serverCsrPEM`/`serverCertificatePEM` mandatory (code 422s neither-present with `INVALID_SERVER_CERT_INPUT`, `registration.go:423-430`), but the edited prose (`api-spec-v2.yaml:181-182`) still reads "can optionally include". One-sentence fix in both mirror files.
- ⚙ **New length checks count bytes; OpenAPI `maxLength` counts code points.** A 200-character CJK revocation comment is 600 bytes → rejected at a third of the documented budget (`v1lifecycle.go:157`; `endpoint.go:34,44`). If bytes is deliberate policy, state the unit in the spec; else `utf8.RuneCountInString`.
- **Replay of a pre-PR-sealed event with a now-dropped reason token now 422s instead of returning the duplicate answer** (ingest validates before dedup, `internal/tl/service/log.go`). Dropped as implausible-precondition: requires an external producer × a dropped-token leaf × a lost-ack replay; this repo's RA is provably immune (domain pre-validates the 7-token set before the event builders run).
- **Present-but-empty strings satisfy the new `required`-based oneOf/anyOf but code treats `""` as absent.** Dropped: only degenerate explicitly-empty discriminators slip through; the realistic no-`omitempty` accident sends both keys and is caught by the oneOf itself; `minLength` appears nowhere in the house spec style.
- **`?status=ALL&status=BOGUS` returns 200 while the reverse order 422s** (pre-existing `ALL` break precedes the new switch). Dropped: only schema-invalid requests are affected, and the 200 they get (unfiltered) is what their own `ALL` requested. Trivial restructure if wanted: validate every value, apply ALL after the loop.
- **Status-filter accept arm has zero test coverage** (no test sends a valid non-ALL status through the new switch) and **no boundary-accept tests at 64/64/5/200**. Dropped as future-regression-only — the arm and limits were verified correct today; both are cheap test-hygiene additions.
- **`?status=` (empty) and lowercase `active` now 422.** Verified as fixes, not breaks: pre-PR both produced a silently-empty 200 (`WHERE status IN ('')`), never the documented default-ACTIVE.
- **Tightened AgentFunction/displayName limits vs already-persisted agents.** Verified non-issue: `AgentFunction.Validate → AgentEndpoints.Validate → NewRegistration` has exactly one non-test caller (`RegisterAgent`, new registrations only); store rehydration is raw struct/JSON assembly; TL events carry no functions/tags.
- **New TL reason-code check reaching read/verify paths.** Verified non-issue: every `Validate()` caller is on the ingest/append path; `statustoken.go`, `identitybadge.go`, `envelope_wrapper.go`, and `cmd/ans-verify` unmarshal without validating — historical leaves keep serving and verifying.
- **`dnssecVerified` schema addition breaking old-leaf validation.** Verified non-issue: field is `omitempty` (absent when false); V2.json's DNSRecord has no `required` list and no `additionalProperties: false`; V1's map-typed attestations structurally can't carry it, so V1.json correctly wasn't touched.
- **Pre-existing V1 envelope `Validate` weaker than V2** (no keyId/signature check at envelope level). Pre-existing asymmetry untouched by this PR; out of scope.

---

## Strengths worth preserving

1. **Docsui mirrors byte-identical to the spec files** (blob 297f598 / d4ffa5b) — issue #59 called for this explicitly.
2. **Symmetric TL ingest guards with a single enum source of truth** (`tl/event/event.go:283`, `tl/event/v1/event.go:191`, both delegating to `domain.RevocationReason.IsValid()`).
3. **Empty-reason passthrough scoped exactly to the issue's ask** (`!= ""` short-circuit + `omitempty`) — non-revocation events unaffected.
4. **Output canonicalized, input leniency retained** (`ParseProtocol`/`ParseTransport` still accept hyphenated forms, new `json_rpc` test) — no break for existing senders.
5. **Status-filter guard matches the spec enum exactly** and preserves ALL/default semantics, with tests documenting the invalid-vs-unfilterable distinction.
6. **The enum-decision thread executed cleanly**: 11-token expansion reverted across all three served schemas in one commit, and the now-inaccurate "RFC 5280" citation dropped.
7. **Precise new spec prose**: the endpoints-uniqueness description names both error codes in the actual check order (`DUPLICATE_ENDPOINT` before `DUPLICATE_PROTOCOL`, matching `endpoint.go:199-211`) and honestly notes JSON-Schema can't express it.
8. **`displayName` required check placed in the domain** (`agent.go:121-123`) so both lanes inherit it — the finding above is what happens when the alternative placement is chosen.

---

**Final count:** 1 finding (0 Critical, 1 Major, 0 Minor). 9 initial → 6 verified → 1 after cross-examination and independent fresh review. A disciplined alignment PR that closes 8½ of 9 drift items — finish the ninth on the lane that declares it, refresh the description, and it's ready.

internal/ra/handler/v1lifecycle.go
jhateley-godaddy: **Served historical schemas narrowed while the spec still claims they're verbatim production fetches** 
  -fixed in https://github.com/agentnameservice/ans/pull/68/changes/20acee641ac199967133ceb40548d9a9ebfa87ed
**Register description prose still says server-cert input is "optional"**
  -fixed in https://github.com/agentnameservice/ans/pull/68/changes/275391163b78e4c627c8d896f0295298aa84160b
**Status-filter accept arm has zero test coverage (no test sends a valid non-ALL status through the new switch)**
  -fixed in https://github.com/agentnameservice/ans/pull/68/changes/938d0d2a4b276e641380a382f44d927800962569
**guard needs a twin on the V2 lane (or a home in the shared service)**
  -fixed in https://github.com/agentnameservice/ans/pull/68/changes/a2af16fb49b5ada36d4c472bc22f82bb86935f61

csnitker-godaddy (approved):
# Re-review of head `e34fb89` — all findings resolved. Approving. ✅

Every item from the previous review was addressed, including three that were only flagged as discretionary:

- **`a2af16f` — the Major finding, fixed.** `LifecycleHandler.Revoke` now enforces `comments ≤ 200` after the `MISSING_REASON` guard with the same `COMMENTS_TOO_LONG` code as V1 and the shared package constant, plus `TestRevoke_CommentsTooLong_422` on the V2 route. Both lanes now enforce the spec's `maxLength: 200`; issue #59 item 9 is fully closed.
- **`20acee6` — historical schemas restored the clean way.** V0.json/V1.json carry the original 11-token enum again; the net change vs main is one description line each (dropping the "RFC 5280" citation — correct, since `EXPIRED_CERT` was never in that RFC). The "production schemas fetched from…" prose is accurate again up to that annotation line, and the historical validation surface is untouched.
- **`2753911` — register prose fixed** ("optionally" → "must"), applied to both mirror files.
- **`938d0d2` — beyond the ask:** all four filterable statuses tested individually plus a multi-value combination test.

**Merged-main interaction checks (the riskier half of this re-review) — all clean:**

- docsui mirrors remain byte-identical to the spec files (`spec/api-spec-v2.yaml` ≡ `docsui/ra.yaml`, blob `5c6cdbc`) even though main's #47 added ~444 lines to the same file and the merge had to reconcile both sides.
- `domain.NewRegistration` still has exactly one non-test caller, so the round-1 verdict that the tightened limits can't retro-break persisted agents survives #47's registration-path changes.
- TL `Validate()` call sites are the identical ingest-only set — #46's event feed and the new `tlclient` adapter added no read-path validation; historical leaves keep serving and verifying.
- No new `RegistrationStatus` constants on main; the status-filter switch remains exactly co-extensive with the spec enum.
- CI green on `e34fb89` (Build & Test, Lint).

**One non-blocking ask:** please refresh the PR description before merging — it still says the RevocationReason enum was *expanded* (`UNSPECIFIED`, `CA_COMPROMISE`, `REMOVE_FROM_CRL`, `EXPIRED_CERT`), but after `a7d3353` + `20acee6` the net schema change is a one-line description edit, so that bullet would mislead anyone reading the merge commit.

Previously-documented discretionary items left as-is (fine to defer): byte-vs-rune length units (now at least consistent across lanes), boundary-accept tests at 64/64/5/200, the `ALL`-order validation skip, and the inherent V1.json-advertises-11 / V1-ingest-enforces-7 tension for new external producers — a one-sentence note in that schema's description would square it whenever the file is next touched.

PR #89, opened 2026-07-21
feat(ra): default discovery profiles to ANS_DNSAID; real-DNS demo tooling

### LINKED ISSUE #86: Set ANS_DNSAID as default DNS Discovery Profile
Once the issue lands for supporting Discovery Profiles in the SDKs -  https://github.com/agentnameservice/ans-sdk-go/issues/56 for example.  We need to circle back and make ANS_DNSAID default.

Make sure the docs in github.com/agentnameregistry/ans-registry are also updated.

csnitker-godaddy:
Closes #86

## What

Flips the V2 default discovery profile set from `{ANS_TXT}` to `{ANS_DNSAID}`: registrations that omit `discoveryProfiles` now provision the DNS-AID-aligned SVCB record family, and the SVCB rows carry the required signal (they are only `Required=false` when ANS_TXT is also requested, per the §4.4.2 transition rule). Operators with legacy `_ans` zone tooling opt into `ANS_TXT` explicitly (or publish the union).

Blast radius is deliberately narrow:

- **V1 lane unaffected** — it never consults `DefaultDiscoveryProfiles()`; it stays pinned to ANS_TXT in `applyDiscoveryProfiles` (covered by `v1_pins_to_ans_txt_ignoring_request_field`).
- **Existing agents unaffected** — profiles persist per-agent at registration (migration 010), so previously registered rows keep the family they were told to publish.

## Shape diff (spec/api-spec-v2.yaml, `AgentRegistrationRequest.discoveryProfiles` + `DiscoveryProfile` enum)

No field names, types, or status codes change. The only wire-visible change is the server-side normalization default for an omitted/empty `discoveryProfiles`:

| | before | after |
|---|---|---|
| `default:` | `["ANS_TXT"]` | `["ANS_DNSAID"]` |
| `example:` | `["ANS_TXT"]` | `["ANS_DNSAID"]` |
| enum prose | ANS_DNSAID "opt-in", ANS_TXT "server default" | ANS_DNSAID "server default", ANS_TXT "opt-in" |

Enforcement site: `internal/domain/dnsrecords.go DefaultDiscoveryProfiles()`. The docsui mirror (`internal/adapter/docsui/openapi/ra.yaml`) is a byte-copy of the spec and updated in lockstep.

## Real-DNS demo tooling

- **`scripts/demo/dns-records.sh` (new)** — operator-facing companion to `register.sh --register-only`: prints the records to publish at each stage (domain-control challenge during PENDING_VALIDATION; production SVCB/badge/TLSA set during PENDING_DNS) with copy-pasteable dig hints, resolves agents by UUID / bare host / `ans://` name, and `--verify` drives `POST verify-dns` through the RA's configured verifier with MISSING/INCORRECT rendering. Dig hints use `TYPE64`/`TYPE65` generic syntax because BIND 9.10's dig (stock macOS) silently queries type A when handed the `SVCB` mnemonic.
- **`start.sh`** — documents `ANS_DNS_TYPE=lookup ANS_DNS_SERVER=1.1.1.1:53` for real-DNS runs and persists the verifier mode to `data/demo/env` for downstream scripts. `common.sh` gains an `ANS_DEMO_SKIP_ENV` guard so that file is strictly start.sh's *output* — a previous run's DNS mode can no longer leak into a fresh stack.
- **`register.sh`** — sends `metaDataUrl` (so ANS_DNSAID SVCB rows carry the `key65400` cap locator) and honors `ANS_DISCOVERY_PROFILES` for explicit profile selection.
- **Fixes found along the way**: `verify-dns.sh` rendered `incorrectRecords` fields as nulls (the DTO nests the expected record under `.record` with the live value in `.found`); `run-lifecycle-v1.sh` never worked under `--with-dns` because it lacked the ACME-challenge-publish step the V2 script has.
- `CLAUDE.md`: lookup verifier record types now include SVCB.

## Verification

- **Live SVCB against 1.1.1.1** through the production `LookupVerifier` (probe against Cloudflare's DDR record `_dns.resolver.arpa`): query, parse, and RFC 9460 §8 subset-match all pass — extra SvcParams (`ipv4hint`, `dohpath`, …) tolerated as designed.
- **Full local E2E** (`start.sh --with-dns`): register → challenge publish → verify-acme → SVCB set printed → negative `--verify` lists MISSING SVCB/badge → `ans-dns install` → positive `--verify` → **ACTIVE**, with the SVCB row sealed into the TL attestation as `dnsRecordsProvisioned[]` `type=SVCB`, `dnssecVerified=true`.
- **Real-DNS negative path**: RA restarted against `1.1.1.1:53` with `--keep`; `verify-dns` queried Cloudflare and returned the exact records to create; `verify-acme` correctly refuses until a challenge artifact is live.
- **Regressions**: `run-lifecycle.sh` and `run-lifecycle-v1.sh` fully green; `make check` passes (coverage 90.7%, gate 90%).

## Notes for reviewers

- Issue #86 also asks for a docs update in `agentnameregistry/ans-registry` — that's cross-repo and not covered here.
- If a union default `{ANS_DNSAID, ANS_TXT}` is preferred for migration safety instead: note the §4.4.2 rule would then mark every SVCB row `Required=false`, so verify-dns would not actually enforce SVCB presence. Sole-DNSAID is what makes the new default meaningful.


jhateley-godaddy:
— Stale DTO comment in registration handler
internal/ra/handler/registration.go Current code (from source, L45-51):

	// DiscoveryProfiles is the set of DNS record families the RA emits
	// for this registration. Each element is one of "ANS_DNSAID" or
	// "ANS_TXT". Typical values: ["ANS_TXT"] (default), ["ANS_DNSAID"],
	// or ["ANS_DNSAID", "ANS_TXT"] (transition union).
	// Empty/missing → ["ANS_TXT"]. Any invalid element rejected
	// with 422 INVALID_DISCOVERY_PROFILE. See ANS_SPEC.md §4.4.2.
	DiscoveryProfiles []string `json:"discoveryProfiles,omitempty"`

scripts/demo/start.sh
```diff
@@ -130,6 +141,12 @@ fi
 # Record the issuer mode for the --keep guard above on the next run.
 echo "$([ "$WITH_ACME" -eq 1 ] && echo acme || echo self)" >"$DATA/issuer-mode"
 
+# Reset the per-run env hand-off file. common.sh sources it in every
+# demo script, so whatever this run appends below (ans-dns zone path,
+# DNS verifier mode/server) is what downstream scripts see — stale
+# values from a previous run must not leak into this one.
+: >"$DATA/env"
```
kperry-godaddy, line 159: This truncate runs before the port-collision refusal at line 151, so re-running start.sh while a stack is up (the `--keep` restart flow this PR's own verification uses) empties the env file the *running* stack's scripts depend on: lifecycle zone publication silently no-ops without `ANS_DNS_ZONE`, verify-dns then fails two steps away, and dns-records.sh reports the verifier mode as "probably noop" while the RA is actually in lookup mode. Pre-PR, the env file was only written inside the `--with-dns` branch, after all the guards. Moving this line (and ideally the issuer-mode write at 142) below the port check restores "a refused start changes nothing" — one-line move.

scripts/demo/verify-dns.sh
```diff
@@ -81,10 +81,13 @@ fi
 RSTATUS=$(printf '%s' "$RESP" | jq -r '.status // empty')
 if [ "$RSTATUS" = "ERROR" ]; then
   # 422 DnsVerificationError — required records aren't live (lookup DNS).
+  # missingRecords is a flat dnsRecord list; incorrectRecords nests the
+  # expected record under .record with the live value in .found.
   warn "required DNS records are not yet satisfied — publish these and re-run:"
   printf '%s' "$RESP" | jq -r '
     (.missingRecords[]?   | "  MISSING    \(.type) \(.name) = \(.value)"),
-    (.incorrectRecords[]? | "  INCORRECT  \(.type) \(.name) (expected \(.value))")' >&2
+    (.incorrectRecords[]? | "  INCORRECT  \(.record.type) \(.record.name) (expected \(.expected); live \(.found))")' >&2
```
kperry-godaddy: The nested `.record` shape this fix targets is V2-only — the V1 422 DTO is flat (`v1DNSMismatchDTO{name, type, expected, found}`, `internal/ra/handler/v1lifecycle.go:84-89`), so on the `--v1` lane this prints `INCORRECT  null null (...)` — the same null-rendering class the PR fixes, moved to the other lane. It's newly reachable because run-lifecycle-v1.sh now works under `--with-dns`. One `//` fallback per field covers both lanes:

```suggestion
  printf '%s' "$RESP" | jq -r '
    (.missingRecords[]?   | "  MISSING    \(.type) \(.name) = \(.value)"),
    (.incorrectRecords[]? | "  INCORRECT  \(.record.type // .type) \(.record.name // .name) (expected \(.expected); live \(.found))")' >&2
```

(The comment on lines 84-85 describes the nested shape as if it applied to both lanes — worth a "V2 nests under .record; V1 is flat" tweak while here.)

scripts/demo/dns-records.sh
```diff
@@ -0,0 +1,266 @@
...
+  fail "verify-dns failed ($ERR_CODE): $(printf '%s' "$RESP" | jq -r '.detail // empty')"
+fi
+RSTATUS=$(printf '%s' "$RESP" | jq -r '.status // empty')
+if [ "$RSTATUS" = "ERROR" ]; then
+  warn "required records are not yet live — publish/fix these and re-run:"
+  # missingRecords is a flat dnsRecord list; incorrectRecords nests the
+  # expected record under .record with the live value in .found.
+  printf '%s' "$RESP" | jq -r '
+    (.missingRecords[]?   | "  MISSING    \(.type) \(.name) = \(.value)"),
+    (.incorrectRecords[]? | "  INCORRECT  \(.record.type) \(.record.name) (expected \(.expected); live \(.found))")' >&2
```
kperry-godaddy: Same V1-lane null rendering as verify-dns.sh:87-89 — this jq is a byte-copy, which is also the maintenance point: the 422 shape now lives in two scripts, and this PR itself shows how the copies drift (the rendering fix landed V2-correct in both places). Shape-tolerant version:

```suggestion
  printf '%s' "$RESP" | jq -r '
    (.missingRecords[]?   | "  MISSING    \(.type) \(.name) = \(.value)"),
    (.incorrectRecords[]? | "  INCORRECT  \(.record.type // .type) \(.record.name // .name) (expected \(.expected); live \(.found))")' >&2
```

Longer term: `--verify` here re-implements verify-dns.sh's whole POST/render/confirm tail — `exec "$SCRIPT_DIR/verify-dns.sh"` with the lane/agent args (or a shared render helper in common.sh) keeps one copy of the DTO knowledge.

scripts/demo/dns-records.sh
```diff
@@ -0,0 +1,266 @@
...
+        "DNS-01   \(.dnsRecord.type) record\n  name   \(.dnsRecord.name)\n  value  \(.dnsRecord.value)\n"
+      elif .type == "HTTP_01" then
+        "HTTP-01  serve at  http://\($host)\(.httpPath // "/.well-known/acme-challenge/\(.token)")\n  body   \(.keyAuthorization // .token)\n"
+      else
+        "\(.type)  token=\(.token)"
+      end'
+    if [ "$LANE" = "v2" ]; then
+      note "then: scripts/demo/acme-verify.sh   (issues certs → PENDING_DNS; re-run this script for the production records)"
+    else
+      note "then: curl -X POST $RA_URL$AGENT_BASE/$AGENT/verify-acme   (→ PENDING_DNS; re-run this script for the production records)"
```
kperry-godaddy: This hint 401s as pasted — verify-acme sits behind the auth middleware like every non-admin route (`cmd/ans-ra/main.go:282`, route at `:367`), and the hint omits the Authorization header the script itself sends everywhere else:

```suggestion
      note "then: curl -X POST -H \"Authorization: Bearer $RA_API_KEY\" $RA_URL$AGENT_BASE/$AGENT/verify-acme   (→ PENDING_DNS; re-run this script for the production records)"
```

(Expanding the key matches how `$RA_URL`/`$AGENT` are already expanded in the hint, and it's the demo key.)

scripts/demo/dns-records.sh
```diff
@@ -0,0 +1,266 @@
...
+
+  note "resolving $input via GET $AGENT_BASE?agentHost=$host&status=ALL"
+  local list matches count
+  list=$(curl -sS -H "Authorization: Bearer $RA_API_KEY" \
+    "$RA_URL$AGENT_BASE?agentHost=$host&status=ALL")
+  matches=$(printf '%s' "$list" | jq -c --arg v "$ver" \
+    '[.items[]? | select($v == "" or .version == $v)]')
+  count=$(printf '%s' "$matches" | jq 'length')
+  case "$count" in
+    0) fail "no registration found for host '$host'${ver:+ version $ver} (owned by this API key)" ;;
```
kperry-godaddy, line 120: Not a blocker: this list call is the one fetch in the script without a problem-document check (the detail GET at 140-141 has one), so a wrong `RA_API_KEY` or a 500 lands in the `0)` arm as "no registration found for host 'X' (owned by this API key)" — a confident misdiagnosis for an auth failure. Same check as the detail fetch:

```sh
err_code=$(printf '%s' "$list" | jq -r '.code // empty')
[ -z "$err_code" ] || fail "agent list failed ($err_code): $(printf '%s' "$list" | jq -r '.detail // empty')"
```

spec/api-spec-v2.yaml
```diff
@@ -1935,15 +1935,15 @@ components:
...
-            the stable default ["ANS_TXT"] server-side; opt into
-            ["ANS_DNSAID"] explicitly. The `minItems`/`uniqueItems`
+            the default ["ANS_DNSAID"] server-side; opt into the legacy
+            ["ANS_TXT"] shape explicitly. The `minItems`/`uniqueItems`
             schema constraints are the canonical client contract for a
             present array — validate before sending. A non-conformant
             request (an explicit empty array, or duplicate values) is
             handled defensively rather than rejected: empty is treated
             as omitted and duplicates are ignored, but conformant
             clients never send either.
```
kperry-godaddy, line 1945: The §4.4.2 required-signal rule is what makes this default meaningful — sole-DNSAID SVCB rows come back `required: true` and verify-dns enforces their presence; add ANS_TXT and every SVCB row flips to `required: false` and is *not* enforced — but the contract never states it (`DnsRecord.required` has no description either). An operator sending the union "for migration safety" reasonably believes SVCB is being verified when it isn't, and first finds out at cutover. One paragraph closes it — and "legacy" reads as the deprecation signal the ANS_TXT enum entry explicitly disclaims ("supported indefinitely"):

```suggestion
            Optional. Omitted (or an explicit empty array) normalizes to
            the default ["ANS_DNSAID"] server-side; opt into the
            ["ANS_TXT"] shape explicitly. The `minItems`/`uniqueItems`
            schema constraints are the canonical client contract for a
            present array — validate before sending. A non-conformant
            request (an explicit empty array, or duplicate values) is
            handled defensively rather than rejected: empty is treated
            as omitted and duplicates are ignored, but conformant
            clients never send either.

            Required-signal semantics (§4.4.2 transition rule): when
            ANS_DNSAID is the sole resolved profile, its SVCB rows are
            returned required: true and verify-dns enforces their
            presence; when ANS_TXT is also requested, the `_ans` TXT
            rows carry the required signal and SVCB rows are returned
            required: false — union registrations do not have SVCB
            presence enforced until ANS_TXT is dropped.
```

Applies to the docsui mirror too (`make docs-sync` handles it).

kperry-godaddy (commented):
Ran a multi-perspective review pass over this PR (security, architecture, systems, Go, skeptic — RA internals and demo scripts reviewed separately). The flip itself verified very clean: single-sourced one-line change, V1 lane pinned in code and test, migration-010 insulation confirmed end-to-end, the §4.4.2 required-signal matrix fully test-pinned, spec/docsui mirror byte-identical. 6 notes inline (4 with committable suggestions), plus these outside the diff:

### Outside this diff
- Recovery for a wedged registration permanently burns the name+version: `ExistsByAnsName` (`internal/adapter/store/sqlite/agent.go:328`) is an unfiltered `COUNT(1)`, so revoke → re-register — the only path for a V2 client whose DNS tooling can't publish SVCB — always forces a version bump, and neither the verify-dns 422 nor the PENDING_DNS nextSteps mentions the `["ANS_TXT"]` opt-out. A remediation sentence in the 422/nextSteps is cheap; a profiles-update endpoint for PENDING_DNS rows is the real fix if wedges show up in practice.
- Nothing at INFO distinguishes a defaulted registration from an explicit `["ANS_DNSAID"]` one (`applyDiscoveryProfiles` normalizes both to the same stored value, and `RegisterAgent` logs nothing on success), so "how many clients did the flip affect since deploy" has no answerable query. One INFO line with a `profilesDefaulted` bool covers it — and fixes registration success being invisible in production logs generally.
- Process/sequencing: issue #86 gated this flip on SDK discovery-profile support — that support is the two open SDK PRs (agentnameservice/ans-sdk-go#63, agentnameservice/ans-sdk-rust#88), whose docs in turn describe this PR's default as already live. Landing all three in one window (SDKs first) satisfies the gate and makes their docs true. "Closes #86" also auto-closes the ans-registry docs deliverable the description defers — a filed follow-up or "Refs #86" keeps it tracked. And since a spec-conformant client's obligations change, `feat(ra)!` with a BREAKING CHANGE footer would surface it in the changelog rather than as an ordinary feature line.
- `internal/ra/handler/registration.go:50` cites "ANS_SPEC.md §4.4.2" — no such file exists in this repo, and that rule is now the justification for the default's enforcement semantics; worth pointing the citation at wherever the rule actually lives.

kperry-godaddy (approved):
All review feedback verified addressed at 4f9f561 — start.sh guard now precedes every state write (issuer-mode included), lane-tolerant incorrectRecords rendering in both scripts, auth'd curl hint, problem-document check on the list call, and the §4.4.2 required-signal semantics in the spec plus a full DnsRecord.required description, mirror in lockstep. CI green. LGTM. Coordination note per #86's gate: ideally merge the SDK PRs (agentnameservice/ans-sdk-go#63, agentnameservice/ans-sdk-rust#88) in the same window, and file the ans-registry docs follow-up before this auto-closes #86.

PR #98, opened 2026-08-01
demo: compose capability descriptors via metaDataUrl + ANS_DNSAID

### LINKED ISSUE #95: Docs: composing payable agent endpoints via metaDataUrl + ANS_DNSAID (no code change)
## Summary

An ANS-registered agent can already be made **payable** without adding a payment protocol, extending the schema, or touching a line of ANS code. The mechanism is the existing optional `metaDataUrl` on an `A2A`/`MCP` `AgentEndpoint`: point it at an agent descriptor that itself advertises payment (an A2A agent-card exposing x402 tools, or a `/.well-known/x402` manifest), pin it with `metaDataHash`, and the RA publishes a `cap` locator through the default `ANS_DNSAID` discovery profile. A resolver discovers the agent by name, follows the integrity-pinned `cap` to the descriptor, and learns how to pay — all over machinery that ships today.

## What I'd like to contribute

A single **`docs/pr-specs/` design-of-record doc** (matching the `FINDER-ard-discovery-service.md` house style) that writes this composition down with a live, independently verifiable worked example (a real A2A agent-card whose `metaDataHash` a reviewer can reproduce with `curl … | sha256sum`).

**Zero code and zero schema change.** No new `Protocol` enum member, no new `DiscoveryProfile`, no change to the RA/TL/verifier or the wire contract. It documents existing behavior and respects the current constraints (catalog eligibility is A2A/MCP-only; `metaDataUrl` host must equal `agentHost`).

## Open design question (deliberately deferred in the doc)

`ANS_DNSAID` publishes SVCB at the bare agent FQDN (`1 .`) carrying the `cap` locator. An operator can *separately* publish a payments-native discovery leaf under a service prefix (e.g. `_pay._agents.<domain>`) following the DNS-AID service-label convention. The doc frames that operator-published record as **complementary** and explicitly does **not** propose that ANS emit such rows or add a discovery profile for them. Whether a payments service token should ever live in the `ANS_DNSAID` emit path is flagged as an open question for maintainers, not resolved here.

## Ask

Per CONTRIBUTING (issue-first, consensus on new material): does a docs-only contribution of this shape sound welcome before I open the PR? Happy to adjust framing or placement to fit house conventions.

sparkmastergrape:
Adds a small runnable demo showing that an ANS-registered agent can advertise arbitrary capabilities through the existing `AgentEndpoint.metaDataUrl`, pinned with `metaDataHash` and published as a `cap` locator by `ANS_DNSAID`. Zero code, schema, or wire changes.

Fixes #95

## Addresses both points from the issue discussion

**Runnable demo with a committed fixture, not a docs file** — `scripts/demo/payable-endpoints/`, with the example card committed as `testdata/agent-card.json`. As you noted, a hash printed against a live card breaks as soon as that card changes, and drift is indistinguishable from a legitimate new version. The demo asserts against the fixture, so the worked example stays verifiable over time; `--live` fetches the real descriptor and *reports* a mismatch without failing, with that ambiguity spelled out as the teaching point.

**Protocol-neutral framing** — the README leads on `metaDataUrl` composing capability descriptors generally. x402 appears only as the worked example, called out as the case people most often assume needs new protocol machinery.

## The expected SVCB row is recorded output, not a derivation

`testdata/expected-svcb.txt` was captured by running `scripts/demo/start.sh` against `main` at `d8ed4bb`, registering the endpoint, driving it to `PENDING_DNS`, and taking `dns-records.sh --json` verbatim:

```
1 . alpn=a2a port=443 key65400=https://api.dnsofmoney.com/.well-known/agent-card.json key65401=IxM_ZcwgCgb3kAnTtD5l3XQ_t9CpFBCSdKbFtanA1Bs key65402=a2a key65409=agent-card.json
```

`verify.sh` derives `cap-sha256` from the fixture and asserts it against that row, so if the emit path in `internal/adapter/discovery/ans/dnsaid.go` changes shape the demo fails rather than agreeing with its own arithmetic. Reproduction steps are in the README.

Note that `register.sh` does not send `metaDataHash`, so the fixture was produced by POSTing `/v2/ans/agents` directly.

## Run it

```
scripts/demo/payable-endpoints/verify.sh
```

No running stack required — it asserts against the committed fixtures.

---

Per the `pay.agents` question — agreed, keeping that out of scope for its own discussion.

**AI assistance:** `Assisted-by: Claude Code (claude-opus-5)`


scripts/demo/payable-endpoints/verify.sh
```diff
@@ -0,0 +1,185 @@
...
+#
+# Exits 0 when the fixture matches its pin; non-zero on a pin mismatch or a
+# missing dependency.
+
+set -euo pipefail
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+FIXTURE="$SCRIPT_DIR/testdata/agent-card.json"
+EXPECTED_SVCB="$SCRIPT_DIR/testdata/expected-svcb.txt"
+AGENT_HOST="api.dnsofmoney.com"
```
csnitker-godaddy: This hardcodes your production host, and the fixture looks like a snapshot of your real card (--live is already reporting different bytes today). Live modes are a feature here, but the other demos take the host from the runner, register.sh honors AGENT_HOST and defaults to a random example.com name. Would you be open to a neutral fixture, with --live taking the descriptor URL as an argument?

scripts/demo/payable-endpoints/verify.sh
```diff
@@ -0,0 +1,185 @@
...
+LIVE=0
+[ "${1:-}" = "--live" ] && LIVE=1
+
+fail() { printf '\nFAIL: %s\n' "$*" >&2; exit 1; }
+header() { printf '\n\033[1m%s\033[0m\n' "$*"; }
+require_cmd() {
+  command -v "$1" >/dev/null 2>&1 || fail "missing required command: $1"
+}
+
+require_cmd sha256sum
```
csnitker-godaddy: sha256sum isn't on stock macOS, so this dies at the dependency check there. Can you add a shasum -a 256 fallback?

scripts/demo/payable-endpoints/README.md
```diff
@@ -0,0 +1,196 @@
...
+# register the endpoint with the metaDataUrl + metaDataHash shown above
+# (register.sh does not send metaDataHash, so POST /v2/ans/agents directly),
+# then drive it to PENDING_DNS:
+#   POST /v2/ans/agents/{agentId}/verify-acme
+scripts/demo/dns-records.sh --json | jq -r '.[] | select(.type=="SVCB") | .value' \
+  > scripts/demo/payable-endpoints/testdata/expected-svcb.txt
+scripts/demo/stop.sh --clean
+```
+
+Recorded against `upstream/main` at `d8ed4bb` on 2026-08-01.
```
csnitker-godaddy: I'd drop this line, git history already carries the provenance and it goes stale on the first re-record.

csnitker-godaddy (changes requested):
Thanks for the quick turnaround, and incorporating the feedback from the issue. Verified the hashes and the recorded row, it all lines up. Left a few inline asks. One naming thought, your call: capability-descriptors might age better than payable-endpoints as the directory name. 

Kicking off CI now.

scripts/demo/payable-endpoints/verify.sh
sparkmastergrape: Done in 93df02e — the fixture is now a neutral card for agent.example.com (nothing serves it, so the offline assertion can't drift), and --live takes the descriptor URL as an argument: it fetches your card and prints its pin, cap-sha256, and the registration body you'd POST. expected-svcb.txt was re-recorded against a fresh ans-ra run with the new card's pin, same recipe as the README's reproduction section.

scripts/demo/payable-endpoints/verify.sh
sparkmastergrape: Added — sha256_file() prefers sha256sum and falls back to shasum -a 256, with a clear failure if neither exists.

scripts/demo/payable-endpoints/README.md
sparkmastergrape: Dropped.

sparkmastergrape:
All four addressed in 93df02e:

- **Neutral fixture** — the card is now a generic agent at `agent.example.com` (nothing serves it, so the offline assertion stays deterministic), and `expected-svcb.txt` was re-recorded against a fresh `ans-ra` run with the new card's pin — same recipe as the README's "Reproducing the fixtures" section, so the asserted row is still recorded reference output rather than script-derived.
- **`--live` takes the descriptor URL as an argument** and composes a registration for it: pin, `cap-sha256`, and the body you'd POST.
- **macOS** — `sha256sum` falls back to `shasum -a 256`.
- **Provenance line dropped.**

Took the `capability-descriptors` rename too — it does age better. Ready for another CI kick whenever you are.

csnitker-godaddy (approved):
LGTM

csnitker-godaddy:
@sparkmastergrape can you go back and sign the commits? I didn't realize they weren't signed until just now, I can merge when that's done

sparkmastergrape:
Signed. Both commits show Verified now.

One thing on your side: CI on the new commits came back `action_required`, so it needs a maintainer to approve workflows for the fork.

PR #103, opened 2026-08-21
feat(ra/tl): add web bot auth identity

### LINKED ISSUE #102: identity: add web-bot-auth as a verified-identity kind for agents
## Summary

The ANS Registration Authority (RA) can verify `did:web`, `did:key`, and `lei` identities for an agent. This POC adds a fourth kind: `web-bot-auth`. An agent operator can prove control of a web-bot-auth `Signature-Agent` URL and bind it to the agent identity.

## Why

Web-bot-auth (`draft-meunier-webbotauth-httpsig-protocol-02`) is the emerging standard for identifying automated HTTP clients. A bot sends a `Signature-Agent` header that names an `https://` URL. That URL serves a directory of the keys the bot signs its requests with. Many agents already have this identity. When they bind it to their ANS name, a relying party can tie an agent ANS name to the signed traffic it sends, without the agent adopting a DID or an LEI.

## Goal

An agent operator registers a `Signature-Agent` URL as an identity, proves control of it, and the RA seals a verified identity. This is the same register, prove-control, and seal flow the other identity kinds already use.

## What the POC should achieve

- Accept a web-bot-auth `Signature-Agent` URL as an identity value.
- Verify that the operator controls a key the URL directory endorses.
- Seal the verified identity to the Transparency Log, so an offline verifier can check it.
- Support the local quickstart with no external network dependency, and a production path that fetches the real directory.
- Tolerate normal key rotation at the directory.

## Out of scope

- Signing or verifying live HTTP request traffic. This POC covers identity registration only, not the per-request `Signature` header a bot sends.

### LINKED ISSUE #123: tl: prevent duplicate ingestion, recover committed leaves, and derive current certificate status
TL ingestion can commit duplicate leaves or leave its SQLite index behind the committed log. Current-state reads can then miss the latest event. Certificate status also diverges between badges and signed status tokens during expiry and renewal overlap.

- Submit the same signed event concurrently. Each request can pass the database content-hash check before either writes its index row, producing multiple envelopes in Tessera.
- Fail the event-index insert after append, then retry or restart. The committed leaf is not recovered before another append.
- Reissue a registration/state with new timestamps. The content hash changes even when the state does not.
- Keep an old certificate in the valid overlap list after a replacement is issued. The old expiry can incorrectly expire the registration.
- Request a status token after the required certificate family expires. Token status and fingerprints can disagree with the badge/current certificate validity.

The fix should combine canonical-content checks, Tessera antispam, serialized lifecycle checks, one POSIX writer, and recovery of all committed leaves before writes and current-state reads. Historical duplicate leaves must remain representable in the index: a unique content-hash constraint otherwise prevents upgrading a log already affected by the race.

Use `min(max(server notAfter), max(identity notAfter))` across required certificate families. Exclude expired fingerprints from current tokens and cap token lifetime at the earliest expiry among the certificates it authorizes.

Preserve the wire shapes, signed bytes, and Merkle history. Expiry semantics are proposed in agentnameservice/ans-registry#68; semantic lifecycle checks also need a corresponding ANS-4 prose clarification.

jhateley-godaddy:
## Related issue

Fixes #102 

## Summary

This branch adds a fourth verified-identity kind to the ANS
Registration Authority (RA): `web-bot-auth`. An agent operator can now
prove control of an `https://` Signature-Agent URL and bind it to the
agent identity, next to the `did:web`, `did:key`, and `lei` kinds.

Web-bot-auth is the IETF HTTP Message Signatures profile for automated
HTTP clients (`draft-meunier-webbotauth-httpsig-protocol-02`). A bot
sends a `Signature-Agent` header that names an `https://` URL. That URL
serves a JWKS directory of the keys the bot signs its requests with.
The URL is the identity. The directory is the authoritative key source.

Lets an ANS agent register that same Signature-Agent URL as a
verified identity. A relying party can then tie the agent ANS name to
the automated traffic it sends.

## Testing

- `internal/crypto/jwk_test.go` — thumbprint vectors and rejection of
  non-Ed25519 / malformed keys.
- `internal/domain/identity_test.go` — inference and canonicalization,
  including the reject paths.
- `internal/adapter/directory/webbotauth/webbotauth_test.go` — the
  parse pipeline, content-type and size checks, key windows.
- `internal/adapter/securefetch/securefetch_test.go` — the dialer and
  redirect guards.
- `internal/ra/service/webbotauthverifier_test.go` — possession,
  endorsement, intersection, mid-rotation drop, alg confusion.
- `internal/ra/service/webbotauth_flow_test.go` — the full
  register → verify → seal flow on the noop resolver, plus rotation.

e2e tests against noop verifier added in scripts/demo/identity-lifecycle.sh

## Checklist

- [x] The PR title follows [Conventional Commits](https://www.conventionalcommits.org) — release notes are generated from it
- [x] Tests cover the change
- [x] The linked issue above uses a closing keyword
- [x] Every commit is signed off (`git commit -s`) certifying the [DCO](https://developercertificate.org)

### DIFF
--- a/cmd/ans-ra/main.go
+++ b/cmd/ans-ra/main.go
@@ -29,6 +29,7 @@ import (
 	"github.com/agentnameservice/ans/internal/adapter/cert"
 	"github.com/agentnameservice/ans/internal/adapter/challenge"
 	"github.com/agentnameservice/ans/internal/adapter/didresolver"
+	"github.com/agentnameservice/ans/internal/adapter/directory/webbotauth"
 	"github.com/agentnameservice/ans/internal/adapter/dns"
 	"github.com/agentnameservice/ans/internal/adapter/docsui"
 	"github.com/agentnameservice/ans/internal/adapter/eventbus"
@@ -160,9 +161,14 @@ func run(cfgPath string) error {
 	// the lei kind's analog of
 	// the did:web resolver selection.
 	leiVerifier := selectLEIVerifier(cfg, logger)
+	// web-bot-auth directory resolver — noop (quickstart) or hardened
+	// HTTPS fetch of the HTTP Message Signatures directory, the
+	// web-bot-auth kind's analog of the did:web resolver selection.
+	dirResolver := selectWebBotAuthDirectoryResolver(cfg, logger)
 	logger.Info().
 		Str("resolver", cfg.Identity.Resolver.Type).
 		Str("vleiVerifier", cfg.VLEI.Type).
+		Str("webBotAuthDirectory", cfg.WebBotAuth.Directory.Type).
 		Dur("challengeTTL", cfg.Identity.ChallengeTTL).
 		Msg("verified-identity surface configured")
 
@@ -247,7 +253,7 @@ func run(cfgPath string) error {
 		identitySealer = tlSealer
 	}
 	identitySvc := service.NewIdentityService(
-		identityStore, identityLinks, agents, didResolver, identitySealer, leiVerifier, db,
+		identityStore, identityLinks, agents, didResolver, identitySealer, leiVerifier, dirResolver, db,
 	).WithSigner(service.EventSigner{
 		KeyManager: km,
 		KeyID:      signerKeyID,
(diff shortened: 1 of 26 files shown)


scourtney-godaddy:
Two notes from verifying the Web Bot Auth path against a live ANS deployment (agent.webmesh.ai):

1. **Key-stability property worth calling out.** An operator's Ed25519 directory key is independent of the ANS identity cert (the cert is backed by a separate CSR keypair) and, in the reference deployment, is pinned to a fixed per-agent key file. So the sealed `web-bot-auth` identity's `keyid` (the RFC 7638 thumbprint) survives cert renewal and re-versioning. That durability is a good property for this kind: a relying party, or a Cloudflare-style directory registration keyed on the thumbprint, can rely on a stable `keyid` across the agent's lifecycle.

2. **Discovery gap for follow-on (belongs to the #102 design, not necessarily this PR).** Once an agent binds a `web-bot-auth` identity, there is no sanctioned ANS surface that advertises "this agent supports web-bot-auth, and here is its directory URL" for the registry or Trust Index to surface. Today it appears only in the NANDA `agentfacts` policies list. A small Trust Card `webBotAuth` field (directory URL + `alg`) would make the sealed identity discoverable, not just verifiable.

The implementation itself (Ed25519-only, sealing the intersection of possession-proven and directory-endorsed keys, SSRF-hardened resolver) matches how real Web Bot Auth directories work.

scourtney-godaddy:
**BLUF:** the Ed25519-only `web-bot-auth` identity kind is correct and matches how live directories behave. It is still draft with no reviews, so #102 stays open until it lands.

Recommended next action: mark it ready for review and merge, which closes #102.

One follow-up for the design, not this PR: nothing in ANS advertises "this agent supports web-bot-auth, and here is its directory URL" for the registry or Trust Index to read. Today that appears only in the NANDA `agentfacts` policies list. A Trust Card `webBotAuth` field fixes it. Captured in `gdcorp-engineering/ans-registry-poc` PR #1234, the new Web Bot Auth compatibility profile.

jhateley-godaddy:
Closing this for now.  ANS agents can already use web-bot-auth without a linked identity.  If a use case arises that requires linking the key to an agent then it can be revisited.

PR #106, opened 2026-08-25
fix(ra): attest only the DNS records verification actually found

csnitker-godaddy:
## What was wrong

When an agent activates, the RA signs an `AGENT_REGISTERED` leaf into the
append-only log. Its `attestations.dnsRecordsProvisioned[]` was built by
walking the *expected* record set rather than the verification results, so
optional records the operator never published were signed in as provisioned.

This is reachable on the default profile with the real lookup verifier. The
TLSA rows are `Required: false`, so an operator who publishes the SVCB and
badge records and skips the TLSA still activates, and the leaf claimed a TLSA
was there. DNSSEC does not catch it: the hard-fail arm needs a live record
that disagrees, and authenticated absence is deliberately allowed through.
With `ANS_TXT` requested it widens to the apex HTTPS record, which a zone
with a CNAME at the apex cannot publish at all (RFC 1034 §3.6.2).

## The fix

One filter at one seam. `VerifyDNS` narrows the set before the seal, and
since `sealActivationEvent` feeds both lane builders, V1 and V2 are both
covered with no signature changes.

Required records need no special case. A required record that is missing or
mismatched returns 422 before the seal, so everything required is already
verified by the time the filter runs. The predicate is `Found` alone.

A nil per-record result means the service was built without a verifier (the
test and embedding path — `cmd/ans-ra` always wires one, `noop` included), so
the expected set passes through unchanged. Empty-but-not-nil is the opposite
and attests nothing. Both are pinned by tests.

## Drop causes in the log

Three different conditions reach that point looking identical, and only one is
normal:

- the operator skipped an optional record
- the operator has a record whose value disagrees, which on an unsigned zone
  raises no 422
- the lookup itself failed

That last one mattered more than expected. `LookupVerifier` returns a hard
error only when it cannot resolve a nameserver at all; a SERVFAIL, REFUSED, or
timeout on a single record comes back as a clean not-found and `VerifyRecords`
still reports overall success. TLSA, SVCB and HTTPS are exactly the query
types older resolvers and middleboxes drop while TXT keeps working, so that
fault leaves the required records healthy and quietly narrows a signed leaf.
Nothing else in the RA reads `RecordVerification.Error`, so this log is where
it becomes visible, and it escalates to WARN:

```
{"level":"info","cause":"MISSING","name":"_443._tcp.agent.example.com","type":"TLSA"}
{"level":"warn","cause":"LOOKUP_ERROR","error":"rcode SERVFAIL","name":"_443._tcp.agent.example.com"}
```

Names, types and causes only, never record values. The filter returns the drop
account from the same pass that builds the attested set, so a separate helper
cannot drift and produce a log line that misdescribes the leaf.

## Doc corrections

Four claims the narrowing made false, or that were already false about this
field:

- `ComputeRequiredDNSRecords` no longer emits the same set it computes.
- The V1 revoke envelope's DNS map is tear-down guidance, not a second
  attestation of the activation state. It does no DNS lookup, so there is
  nothing to narrow against, and a verifier should not diff it against the
  registration leaf.
- The `discoveryProfiles` description in `spec/api-spec-v2.yaml` (docsui
  mirror re-synced via `make docs-sync`).
- `event.DNSRecord.DNSSECVerified` claimed TLSA only, while the verifier has
  been setting it for SVCB and HTTPS too.

## Not changing

- **No wire-shape change.** `dnsRecordsProvisioned[]` is already variable
  length by profile and endpoint count, and neither lane's `Validate`
  inspects attestations, so a narrower or empty array cannot be rejected at
  TL ingest.
- **The attested value stays the RA's expected `Value`**, not `r.Actual`, so
  operator-controlled bytes stay out of the signed leaf and a benign
  `Actual != Value` delta on an SVCB subset match does not change the bytes.
- **`dnssecByKey` stays keyed on name+type.** DNSSEC signs RRsets, so records
  sharing an owner name and type share one AD bit by construction.
- **The V1 revoke code itself.** Fixing it properly means either dropping the
  field from the V1 revoke envelope, which breaks reference-schema parity, or
  persisting what was attested at activation, which is a storage change.
  Documented rather than changed.

## Verification

- `make check` passes: fmt, vet, golangci-lint, coverage 90.7% against the
  90% gate.
- `go test -race ./internal/ra/... ./internal/tl/...` clean.
- `recordKey`, `attestedDNSRecords`, `dropCause`, `droppedForLookupError` all
  at 100% statements.
- Ran the real `LookupVerifier` against an in-process DNS server with no TLSA
  published to confirm the premise the tests rest on: `Found=false`,
  `Actual=""`, `AllRequired=true`.
- Mutation-checked the two predicates that carry the correctness:

  | Mutation | Result |
  |---|---|
  | `Found` → `Found \|\| Actual != ""` | fails `optional_with_mismatched_value_is_dropped` |
  | remove the `Error` arm from `dropCause` | fails `optional_lookup_error_is_dropped_and_not_reported_as_missing` |

## Follow-up worth its own PR

TLSA rollover on a DNSSEC-signed zone. Multiple TLSA records at one owner name
work fine today — `verifyTLSA` loops the answer set and returns on the first
match, so extra siblings are ignored. But mid-renewal, before the new record is
published, the old one alone gives `Found=false` with a non-empty `Actual` and
the AD bit set, which trips the hard-fail arm and 422s even though TLSA is
optional. An unsigned operator sails through with the record simply dropped.
Making that gate group-aware (only hard-fail when *no* record at that name
matches, per RFC 7671 §8.1) pairs naturally with adding the `3 1 1` SPKI
variant, since a `3 1 1` record present with no matching `3 0 1` looks
identical to a rewritten one.

### DIFF
--- a/internal/ra/service/dnsrecords.go
+++ b/internal/ra/service/dnsrecords.go
@@ -4,13 +4,17 @@ import (
 	"github.com/rs/zerolog/log"
 
 	"github.com/agentnameservice/ans/internal/domain"
+	"github.com/agentnameservice/ans/internal/port"
 )
 
 // ComputeRequiredDNSRecords returns the DNS records the operator must
 // publish for reg, composed by walking the discovery registry. The RA
 // does not create these records — the operator manages their own DNS;
-// the RA only verifies they exist and emits the same set onto the TL
-// as `dnsRecordsProvisioned[]`.
+// the RA only verifies they exist.
+//
+// This is the asked-for set, not the attested set. What reaches the TL
+// as `dnsRecordsProvisioned[]` is this set narrowed to the records DNS
+// actually answered with; see attestedDNSRecords.
 //
 // Composition rules:
 //
@@ -75,7 +79,7 @@ func (s *RegistrationService) ComputeRequiredDNSRecords(reg *domain.AgentRegistr
 			// false), so first-seen wins. A profile that disagreed
 			// would have its flag silently dropped here — keep the
 			// flags aligned across adapters.
-			key := r.Name + "|" + string(r.Type) + "|" + r.Value
+			key := recordKey(r)
 			if seen[key] {
 				continue
 			}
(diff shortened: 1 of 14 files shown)


internal/ra/service/dnsrecords.go
```diff
@@ -119,6 +123,152 @@ func (s *RegistrationService) ComputeRequiredDNSRecords(reg *domain.AgentRegistr
...
+	case !haveResult:
+		return dropCauseNoResult
+	case v.Error != "":
+		return dropCauseLookupError
+	case v.Actual == "":
+		return dropCauseMissing
+	default:
+		return dropCauseMismatch
+	}
+}
```
kperry-godaddy, line 256: `dropCause` classifies any result carrying a non-empty `Error` as `LOOKUP_ERROR`, and `droppedForLookupError` escalates that drop log to WARN. The lookup adapter, though, reports every non-success rcode through `Error` — including NXDOMAIN, which is what an operator who never created `_443._tcp.<fqdn>` actually gets (RFC 2308 §1: the name doesn't exist at all; checked live, `dig _443._tcp.example.com TLSA` returns NXDOMAIN, and `ans-dns` sets `RcodeNameError` whenever the answer set is empty). So the routine skip-DANE activation this INFO/`MISSING` arm was built for lands in the WARN/`LOOKUP_ERROR` arm instead, and the WARN channel fills with ordinary operator choices from the first real activation — hiding the genuine SERVFAIL/timeout narrowing it exists to surface. The new tests model the skip case as `Error: ""`, which the real verifier only produces for NODATA (name exists, no records of that type), not for a nonexistent name.

The narrowing itself is unaffected — the record is dropped either way — and the required-record path partitions on `Actual`, never `Error`, so the 422 behavior doesn't move. The fix fits in the adapter (`internal/adapter/dns/lookup.go`, all four `verify*` functions): treat NXDOMAIN as authoritative absence so only SERVFAIL/REFUSED/timeouts populate `Error`:

```go
	if resp.Rcode == dns.RcodeNameError {
		// NXDOMAIN: authoritative "this name does not exist" —
		// clean absence, not a lookup fault.
		return r
	}
	if resp.Rcode != dns.RcodeSuccess {
		r.Error = fmt.Sprintf("rcode %s", dns.RcodeToString[resp.Rcode])
		return r
	}
```

Worth fixing before this ships — the INFO/WARN split is the diagnostic this change adds, and it inverts for the most common drop cause on real DNS. An adapter test pinning NXDOMAIN → `Found=false, Error==""`, plus an NXDOMAIN-shaped `MISSING` case in the new table, would lock it in.

kperry-godaddy (commented):
Ran a multi-perspective review pass over this PR (security, architecture, systems/operability, Go idioms, plus an adversarial claims check). The attestation narrowing held up everywhere it was probed: the single filter genuinely covers both lanes, the fail-closed defaults are right, the wire-compat claim verifies against both lanes' `Validate`, and the predicate's justification against the mismatch shape checks out. One finding worth acting on before merge, inline below — it's in the drop-cause classification the change adds, not in the filter.

internal/ra/service/dnsrecords.go
csnitker-godaddy: Confirmed, and it reproduces with our own tooling — `cmd/ans-dns/main.go:201` sets `RcodeNameError` whenever the answer set is empty, so a local run hits it too.

Fixed in the adapter, though I pulled the rcode handling into a single `inspectAnswers` helper rather than repeating the check in all four `verify*` functions. NXDOMAIN returns with `Error` empty and joins NODATA on the MISSING arm; only SERVFAIL, REFUSED and transport failures populate `Error`. `DNSSECVerified` stays false there, as in your snippet — the hard-fail rule needs a live value to disagree with, and a dropped record never reaches the wire, so setting it would be inert. That reasoning is in the doc comment.

On whether it belonged in this branch: `Error` had no behavioural reader before it. The only non-test reads are the two lines this branch adds, so the two tests pinning "NXDOMAIN surfaces an error" were describing adapter internals rather than a contract — which is what made inverting one named `NXDOMAINSurfacedAsError` the right call. `port.RecordVerification.Error` now states which outcomes qualify, since the service layer leans on the split.

Your point about the fixtures is what explains how it got through. The harness could only ever produce NXDOMAIN, because an empty answer set became `RcodeNameError`, so NODATA had no coverage at all and the service fixtures modelled absence as `Error: ""` — agreeing with the intent while the adapter disagreed with both. Hand-built port values can't catch that, since they encode what the author thinks the adapter does.

Three layers now. The harness produces NXDOMAIN, NODATA and SERVFAIL independently, and the adapter pins all three across TXT, TLSA, HTTPS and SVCB. A seam test pins `dropCause` against each shape the adapter emits. And an integration test runs the real verifier against a real nameserver and hands the output to the real filter — that's the one that would have caught this originally, and reintroducing the defect fails it with "declining an optional record is MISSING, not LOOKUP_ERROR".

csnitker-godaddy:
Addressed Keith's finding. The diff now reaches past the RA service into `internal/adapter/dns/lookup.go` and `internal/port/dns.go`, which wasn't in the original scope. Short version: this branch is what first gave `port.RecordVerification.Error` behavioural meaning, and the adapter was populating it for NXDOMAIN — which is exactly what an operator who declines an optional record produces. Reasoning on the thread.

internal/ra/service/dnsrecords.go
kperry-godaddy: Verified at e526524. The helper reads cleanly and does what this thread needed — NXDOMAIN joins NODATA on the empty-`Error` arm, so declining an optional record lands on MISSING/INFO, and only SERVFAIL/REFUSED/transport failures escalate. Agreed on `DNSSECVerified` staying false there: the hard-fail rule needs a live value to disagree with, and a dropped record never reaches the wire, so the bit would be inert. Inverting `NXDOMAINSurfacedAsError` was the right call for the reason you gave — before this branch nothing read `Error`, so the old assertion pinned an internal, not a contract.

The test response is stronger than what I sketched: the harness producing NXDOMAIN, NODATA, and SERVFAIL independently closes the gap that let this through, and the integration test — real verifier, real nameserver, real filter — is the one that catches this class of disagreement between self-consistent units. Ran both packages with the race detector locally: clean. Resolving.

kperry-godaddy (approved):
The one issue from the review pass is fixed and verified at e526524: NXDOMAIN now classifies as absence rather than a lookup fault, so the drop log's INFO/WARN split holds on real DNS, and the new adapter, seam, and integration tests pin the behavior end to end. vet, race detector, and CI are green. The remaining review notes are follow-up material, none blocking.

PR #108, opened 2026-08-29
fix(dns): compare SVCB TargetName as effective target, not presentation string (RFC 9460 §2.5.2)

martian-pink:
`matchesSVCBSubset` compared the SVCB TargetName as a raw presentation string. The RA builds the expected row with target `.`, so a live record whose target is the explicit owner FQDN (the form emitted by publishing tools such as the DNS-AID reference implementation) failed `verify-dns` with "record not found", despite designating the same target per RFC 9460 §2.5.2:

> For ServiceMode SVCB RRs, if TargetName has the value ".", then the owner name of this record MUST be used as the effective TargetName.

This normalizes both sides to the effective TargetName before comparing: in ServiceMode `.` expands to the record's owner name, and both targets are canonicalized (lowercase, trailing dot) so the comparison is case- and FQDN-dot-insensitive, as DNS names are. AliasMode (priority 0) `.` keeps its "service not available" meaning and is left as-is.

Assisted-by: Claude Code (claude-fable-5)

### DIFF
--- a/internal/adapter/dns/dns_test.go
+++ b/internal/adapter/dns/dns_test.go
@@ -304,6 +304,34 @@ func TestLookupVerifier_SVCB(t *testing.T) {
 			why:       "ServiceMode expectation should not match an AliasMode record",
 		},
 		{
+			// RFC 9460 §2.5.2: in ServiceMode, TargetName "." designates
+			// the record's owner name, so a live record whose target is
+			// the explicit owner FQDN — the form emitted by publishing
+			// tools such as the DNS-AID reference implementation — is
+			// the same target as the expected ".". Comparing the
+			// presentation strings literally would report a present,
+			// correct record as not-found.
+			name:      "servicemode-explicit-owner-fqdn-matches-expected-dot",
+			zoneName:  "agent.example.com.",
+			zoneRR:    `agent.example.com. 3600 IN SVCB 1 agent.example.com. alpn=a2a port=443`,
+			queryName: "agent.example.com",
+			want:      `1 . alpn=a2a port=443`,
+			found:     true,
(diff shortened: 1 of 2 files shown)


kperry-godaddy (commented):
Thanks for this — it's a well-built fix. I verified the RFC 9460 §2.5.2 reading, ran the new table cases (green, race-clean), and checked the premise independently: the DNS-AID reference implementation does publish explicit owner-FQDN targets, so the literal comparison really was rejecting correct records — and in a DNSSEC-signed zone, reporting them as tampering. Normalizing at the comparison site rather than changing the expected value the RA emits was the right call (the expected value is what lands in the transparency-log attestation, so it has to stay byte-stable), and the negative case pinning "a different explicit target still fails" is exactly the guard this needed.

Two items from [CONTRIBUTING.md](https://github.com/agentnameservice/ans/blob/main/CONTRIBUTING.md) before this can merge — both live on the commit and one amend covers them:

- **DCO sign-off** — the red check is the missing `Signed-off-by:` trailer.
- **AI attribution** — the `Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>` trailer has to come off: DCO certification belongs to the human submitter, so AI tools don't take author trailers here. Disclosure lives in the PR description instead, e.g. `Assisted-by: Claude Code (claude-fable-5)`.

`git commit --amend -s` (dropping the trailer in the same edit) and a force-push takes care of both.

kperry-godaddy (approved):

kperry-godaddy (changes requested):
@martian-pink - We just need to have the commits signed

kperry-godaddy (approved):

Issue #99, opened 2026-08-07
ans-verify: add general agent verification (non-ANS agents) and distinguish ans_verified from reachable


scourtney-godaddy:
## Summary

The `ans-verify` CLI verifies ANS TL receipts and Merkle proofs correctly for ANS-registered agents. Today's agent-to-agent interoperability testing revealed two gaps worth tracking:

## 1. No general agent verification for non-ANS agents

The current `ans-verify` tool requires an `agentId` and queries the TL. It cannot verify an agent that is discoverable via DNS-AID SVCB and has a reachable signed agent card but is not registered in ANS.

In practice, legitimate agents from other ecosystems (Infoblox DDI agent, for example) are discoverable via DNS-AID SVCB at the owner FQDN, have signed agent cards with `jku`, and enforce proper OAuth2 auth — but return no ANS TL proof. A caller interoperating with these agents needs a verification path that reports DNS state, SVCB records, DNSSEC, and card reachability independently of ANS registration.

Proposed: extend `ans-verify` (or add an `ans-verify agent <fqdn>` subcommand) that:
1. Queries `_ans` TXT (ANS discovery record)
2. Queries SVCB at the bare FQDN (DNS-AID)
3. Fetches the agent card, trying the SVCB TargetName when the owner name has no A record
4. If an ANS `agentId` is found, runs the existing TL proof check
5. Reports `ans_verified` (TL proof valid) and `reachable` (card fetched) as distinct outputs

## 2. Terminology: `verified` vs `ans_verified`

The current tool exits 0 for "verified." For interoperability, the distinction matters:

- `ans_verified=true` — the ANS TL Merkle proof is valid. Requires ANS registration.
- `reachable=true` — the agent card was fetched and the agent is contactable. Does not require ANS registration.

An agent can be `reachable=true, ans_verified=false` (DNS-AID discoverable, not in ANS) or `ans_verified=true, reachable=false` (TL proof valid but endpoint currently down). Both states are meaningful and different.

Suggesting the output schema use these two fields rather than a single `verified` boolean.

## Reference

Encountered while testing against Infoblox `ddi-agent.ai.infoblox.com` (DNS-AID SVCB, OAuth2 auth, signed card with `jku`). The SVCB for that agent points to a different TargetName (`ddi-agent-edge.ai.infoblox.com`) with no A record at the owner FQDN — a pattern the current tool cannot follow.

csnitker-godaddy:
[...]

PR #112, opened 2026-09-04
feat(ans-verify): verify attested metadata hashes against the served descriptor

sparkmastergrape:
Refs #99 — the part @csnitker-godaddy said belongs in `ans-verify`: the RA attests a metadata hash and nothing checks it.

**What.** `ans-verify` gains Step 7. For every `attestations.metadataHashes[protocol]` in the receipt's leaf it finds the attested SVCB/HTTPS row for that protocol (matched on `alpn=` / `key65402=`), reads the descriptor URL from that row's `key65400=`, fetches it, and compares SHA-256 with the attested value. Both inputs come from the same signed leaf, so the check is closed over what the log carries: no DNS lookup, no operator input. Since #106 the attested rows are the ones verification actually found, so the URL is one the RA saw served.

**Outcomes.** Mismatch → `✗` and exit 1 (the registration asserts something that is not true). Unreachable (network error or non-200, a 401 included) → `⚠`, exit 0. A hash with no attested row carrying a URL → `⚠`, exit 0. No attested hashes → `✓ nothing to check`. `-check-metadata=false` skips the step; `-metadata-timeout` (default 15s) bounds each fetch. Bodies are capped at the existing `maxResponseBytes`.

**Lanes.** Reads the V2 `dnsRecordsProvisioned[]` array and the V1 name→data map.

**Tests.** 14 cases in `metadata_test.go`: V2 pairing incl. HTTP_API→`x-http`, V1 map shape, nothing attested, bad JSON, unknown record shape, bap-only rows and first-row-wins, the four outcomes against an `httptest` server, body cap, bad URL, and the step's fail-only-on-mismatch rule (six sub-cases). The demo `run-lifecycle.sh` registers no metadata hash, so its step 14 prints "nothing to check" and stays green.

**Gate.** `go vet` ok, `golangci-lint` (v2.11.4) 0 issues, `go test -race ./...` clean, coverage 90.7% (the `test-cover` recipe; `cmd/` is outside the gate by design, and every new function sits at 87–100%).

Not in this PR: the FQDN entry point from the issue. Happy to do it next as its own PR.

Assisted-by: Claude Code (claude-fable-5-1)

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,6 +192,11 @@ The verifier:
 6. ES256-verifies the signature over Sig_structure.
 7. Cross-checks that the receipt's computed leaf hash matches the
    badge's `merkleProof.leafHash` (same leaf in the same tree).
+8. For every metadata hash the leaf attests, fetches the descriptor at
+   the URL the attested SVCB row carries (`key65400`) and compares its
+   SHA-256 with the attested value. A mismatch fails verification; an
+   unreachable descriptor is reported only. `-check-metadata=false`
+   skips this step.
 
 ## Configuration
 
--- a/cmd/ans-verify/main.go
+++ b/cmd/ans-verify/main.go
@@ -23,6 +23,12 @@
 //     verification keys.
 //  6. Fetches the badge (/v1/agents/{agentId}) and cross-checks
 //     that the badge's merkleProof matches the receipt's proof.
+//  7. For every metadata hash the leaf attests, fetches the
+//     descriptor at the URL the attested SVCB row carries (key65400)
+//     and compares its SHA-256 with the attested value. A mismatch
+//     fails verification: the registration asserts something that
+//     is not true. An unreachable descriptor is reported only.
+//     Disable with -check-metadata=false.
 //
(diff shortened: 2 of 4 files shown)


csnitker-godaddy (approved):
Thanks for taking this one. I read it against the RA side and it lines up. The leaf has the hash per protocol and the SVCB row has the URL, so the check only uses what's already in the signed leaf. Nothing else looks at that hash today, so this is the first time the claim actually gets tested. Mismatch fails and unreachable warns is what I had in mind. Tests and lint pass for me locally too.

One small thing, not blocking. If the agent has been revoked, the latest leaf is the AGENT_REVOKED one and step 7 says the leaf attests no metadata hash. That's true of that leaf but reads like the registration never declared one. Might be worth a different message there, or skipping the step when the event isn't a registration. Your call.

On #99, this is the hash piece only, so I'd leave the issue open. The FQDN entry point is still the part I'd like to see land. Happy for that to be its own PR like you said.

PR #113, opened 2026-09-08
feat(dns): add TLSA 3 1 1 (SPKI) record builder for auto-renewing certs

### LINKED ISSUE #105: feat(dns): offer TLSA 3 1 1 (SPKI) for auto-renewing certs; 3 0 1 hardcode breaks on rotation
**BLUF:** `TLSARecord` hardcodes selector 0 (full certificate), so the published `3 0 1` value changes on every certificate renewal and must be re-published each time. For a 90-day auto-renewing Let's Encrypt certificate this breaks DANE (DNS-based certificate pinning) on each rotation unless the operator updates DNS at the same time. The profile should offer selector 1 (SubjectPublicKeyInfo, the public key), which stays valid across renewal when the key is reused.

## Where

`internal/adapter/discovery/ans/tlsa.go` emits `3 0 1` (usage 3 DANE-EE, selector 0 full-cert, matching type 1 SHA-256). Its comment states the value changes on every cert rotation.

## DNS-AID direction

DNSAID-02 is selector-agnostic: it references RFC 6698 and RFC 7671 and says TLSA MAY be used. RFC 7671 §5.1 recommends SPKI (`3 1 1`) for most deployments, because it survives certificate reissuance when the key is retained.

## Proposed change

Let the profile emit `3 1 1` (SPKI) as an option, or default to it for endpoints whose certificate renews automatically. An operator that reuses the keypair across renewals then keeps one stable TLSA record rather than re-publishing a full-cert hash every cycle.

## Observed

On `webmesh.ai`, each agent ended up with both a `3 0 1` (the RA-required full-cert value) and a hand-added `3 1 1` (for renewal survival). A single `3 1 1` would suffice and would not go stale on renewal.

scourtney-godaddy:
## BLUF

The reference impl emitted only `TLSA 3 0 1` (full-cert hash), which changes on every renewal and breaks DANE for auto-renewing certs. This adds a `3 1 1` (SPKI-hash) record builder that survives renewals reusing the same key; the existing `3 0 1` builder is untouched.

Closes https://github.com/agentnameservice/ans/issues/105

## What changed

**internal/domain/dnsrecords.go** -- new TLSARecordForCertSPKI(fqdn, spkiFingerprint string) emitting 3 1 1. The existing TLSARecordForCert (3 0 1) is untouched.

**internal/crypto/x509.go** -- two new helpers: SPKIFingerprint(*x509.Certificate) and SPKIFingerprintFromPEM(pem string). Also extracts pemTypeCertificate constant to resolve a goconst finding.

**Tests** -- dnsrecords_test.go asserts rec.Value == "3 1 1 " + hex(sha256(cert.RawSubjectPublicKeyInfo)) using a generated cert. A companion test asserts the SPKI fingerprint differs from the full-cert fingerprint (the selector-1 correctness trap).

## Why 3 1 1 survives rotation

Selector 0 (full cert) hashes the entire DER certificate -- changes on every renewal. Selector 1 (SPKI) hashes only the SubjectPublicKeyInfo -- stable across renewals that reuse the same key pair. This is the property ACME auto-renewing CAs rely on.

## Caller availability finding

All three TLSARecordForCert call sites are in internal/ra/service/renewal.go (lines 222, 372, 475). At those sites the code holds a *domain.ByocServerCertificate (LeafCertificatePEM) or *port.ValidatedCert (LeafPEM) -- no pre-parsed *x509.Certificate and no pre-computed SPKI hash. Wiring up TLSARecordForCertSPKI requires one additional anscrypto.SPKIFingerprintFromPEM call per site. Not a structural refactor; straightforward to wire up in a follow-on or in this PR per reviewer preference.

## Design options (recommendation: a)

**(a) Replace 3 0 1 with 3 1 1 as the default in the renewal-status DTO.** The DTO is an operator hint; the selector change is backward-compatible and benefits auto-renewing BYOC immediately. This is my recommendation.

**(b) Emit both records.** Valid DANE; gives operators a transition window. Doubles DNS records per hostname.

**(c) Make it configurable.** Adds a flag to the registration model. Appropriate only if operators need 3 0 1 long-term (CAs that replace keys on each issuance); not the common BYOC path.

### DIFF
--- a/internal/crypto/x509.go
+++ b/internal/crypto/x509.go
@@ -25,6 +25,9 @@ import (
 // Baseline Requirements for publicly trusted certificates.
 const (
 	MinRSAKeyBits = 2048
+
+	// pemTypeCertificate is the PEM block type header for X.509 certificates.
+	pemTypeCertificate = "CERTIFICATE"
 )
 
 // AllowedSignatureAlgorithms is the allowlist of signature algorithms
@@ -61,7 +64,7 @@ var (
 // ParseCertificatePEM parses a single PEM-encoded X.509 certificate.
 func ParseCertificatePEM(pemData string) (*x509.Certificate, error) {
 	block, _ := pem.Decode([]byte(pemData))
-	if block == nil || block.Type != "CERTIFICATE" {
+	if block == nil || block.Type != pemTypeCertificate {
 		return nil, fmt.Errorf("%w: expected CERTIFICATE block", ErrPEMParse)
 	}
 	cert, err := x509.ParseCertificate(block.Bytes)
@@ -82,7 +85,7 @@ func ParseCertificateChainPEM(pemData string) ([]*x509.Certificate, error) {
 		if block == nil {
 			break
 		}
-		if block.Type != "CERTIFICATE" {
+		if block.Type != pemTypeCertificate {
 			continue
 		}
 		cert, err := x509.ParseCertificate(block.Bytes)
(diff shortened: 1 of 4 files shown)


csnitker-godaddy (changes requested):
Posted the decision on #105. As it stands this PR adds helpers that nothing calls, and the one place it proposes to use them is the renewal DTO, which is not where the record set comes from. The registration emitter, verify-dns, and the sealed record set all go through `ans.TLSARecord`, not `domain.TLSARecordForCert`. If we merge just this, the RA has two TLSA builders that disagree and a set of helpers with no caller. I'd rather not leave it in that state, so the rest needs to land here rather than in a follow-up.

What I'd want in this PR:

1. verify-dns matches any DANE-EE row against the server cert under the row's own selector and matching type, same as #71 and #107, and records the row that matched as Actual. Today it compares one expected string, so a signed zone serving the other form is a hard 422, which is exactly what existing operators would hit once we emit `3 1 1`.
2. The activation leaf seals the row that was found, not the precomputed one.
3. `ans.TLSARecord` and the renewal DTO both emit `3 1 1`. No `3 0 1` going forward.
4. Tests for the above.

With all of that in, "Closes #105" is true. Let's get #71 and #107 released ahead of this so DANE callers have an SDK that accepts the new form. I'll open the spec PR alongside.

PR #116, opened 2026-09-08
fix(dns): match the HTTPS record by SvcParam subset, with alpn as a list (RFC 9460 §7.1.1)

### LINKED ISSUE #115: fix(dns): verify-dns rejects the HTTPS record every CDN-proxied name serves — type 65 is matched verbatim
## Summary

`verifyHTTPS` compares the served HTTPS (type 65) record against the expected `1 . alpn=h2` verbatim — `normalizeHTTPS` collapses whitespace and nothing else. `verifySVCB`, in the same file, has done subset matching since #108; the sibling record type never got it.

A CDN edge synthesizes the HTTPS record for a proxied name and serves it in place of any manually added one, so where the RA expects `1 . alpn=h2` the zone answers `1 . alpn=h3,h2` with `ech` and address hints attached. In a DNSSEC-signed zone the lifecycle layer then reads the disagreement as tampering — `HTTPS_DNSSEC_MISMATCH`, a hard fail regardless of `Required: false` — so the record's optionality never reaches the operators it was written for: the branch that grants the exemption requires the record to be *absent*, and theirs is present with content they do not control.

## Scope

The HTTPS row is emitted by `TXTProfile` alone (`internal/adapter/discovery/ans/txt.go`), gated on `len(reg.Endpoints) > 0`; `DNSAIDProfile` emits SVCB and no HTTPS row. So this affects registrations that resolve `ANS_TXT`: every V1-lane registration — `applyDiscoveryProfiles` (`internal/ra/service/helpers.go`) pins V1 to `[ANS_TXT]`, ignoring what the client sent — and any V2 registration that names the profile.

## Where

- `internal/adapter/dns/lookup.go`, `verifyHTTPS`: matches on `normalizeHTTPS(got) == wantNorm`, where `normalizeHTTPS` is `strings.Join(strings.Fields(s), " ")`.
- `internal/adapter/discovery/ans/txt.go`, `TXTProfile.Records`: expected `1 . alpn=h2`, `Required: false`, at the bare FQDN. The doc comment gives the reason for the flag: *"Required=false because operators on CNAME-fronted apex zones cannot publish this record at the same name (CNAME at @ blocks HTTPS RR per RFC 1034 §3.6.2); the spec does not block them on its absence."*
- `internal/ra/service/lifecycle.go`, `verifyDNSRecords`: `r.DNSSECVerified && !r.Found && r.Actual != ""` on TLSA, SVCB or HTTPS produces `<RECORD_TYPE>_DNSSEC_MISMATCH` and a 422. The comment explains the `Actual != ""` gate precisely: authenticated NODATA is *"authentic ABSENCE, not tampering"*, so an operator who omits an optional record falls through to the Required check and is skipped.

That gate is what a proxied name cannot reach. It is the same operator the `Required=false` comment describes — one who cannot publish the record — but with a value in place of the void.

## Why the operator cannot satisfy it

Cloudflare's DNS record documentation states that proxied names get a synthesized HTTPS record whenever Universal SSL is on, and that *"Manually-added HTTPS records on proxied names are not served"*. That leaves three states, none of which verifies: leave the zone alone and the served value is the provider's; add `1 . alpn=h2` and the record sits in the zone unserved; and no configuration makes the name answer NODATA while it stays proxied.

The two escapes the provider documents are turning every record at that name DNS-only — the edge goes with it, so no proxy, no WAF, no CDN — or disabling Universal SSL for the zone. Both are steep prices for a record the profile marks optional.

## Reproduce

In the repo's own harness, alongside `TestLookupVerifier_HTTPSMatch`:

```go
s.add("agent.example.com.", "HTTPS",
    `agent.example.com. 300 IN HTTPS 1 . alpn="h3,h2" ipv4hint=192.0.2.1 ech="AEX+DQBB" ipv6hint=2001:db8::1`)

recs := []domain.ExpectedDNSRecord{{
    Name: "agent.example.com", Type: domain.DNSRecordHTTPS,
    Value: `1 . alpn=h2`, Required: false,
}}
```

`verifyHTTPS` returns `Found: false` with `Actual` populated. Feed that to `verifyDNSRecords` with `DNSSECVerified: true` and it yields `HTTPS_DNSSEC_MISMATCH`, which the handler maps to 422 — on a record set that is otherwise correct, in a zone whose only distinguishing properties are that it is signed and behind a proxy.

One detail for the assertion: `formatHTTPSValue` renders SvcParams through `miekg/dns`'s `p.String()`, unquoted, so the string under comparison is `1 . alpn=h3,h2 ipv4hint=… ech=… ipv6hint=…`, while `dig` prints the same record with the alpn list quoted.

## Proposed change

Reuse the SVCB matcher for HTTPS: priority equal, effective TargetName equal (RFC 9460 §2.5.2, as in #108), every expected SvcParam present, additional SvcParams tolerated (§8). One refinement SVCB did not need: `alpn` is a list (§7.1.1), so an expected `h2` must match a served `h3,h2` — compare `alpn` as set inclusion, expected ⊆ served, and the remaining params by equality.

The hard-fail path stays intact for real tampering: a different priority, a different effective target, or an `alpn` that does not carry the expected protocol.

## Why this is worth fixing rather than documenting

1. **The optional flag does not protect the operator it was written for**, per the mechanism above.
2. **The shape is not niche, and it moves.** Every proxied name with Universal SSL gets a synthesized HTTPS record, and the `ech` and hint params inside it rotate under the operator's feet — so a verbatim match is unstable even where it passes once, which is how a registration can verify one week and fail the next with nothing changed on the operator's side.
3. **SVCB already went through this.** #108 fixed presentation form being treated as identity in the sibling record type. This is the same matcher problem with one list-valued param added.

PR follows.

martian-pink:
Fixes #115.

**BLUF.** `verifyHTTPS` compared the served record against the expected `1 . alpn=h2` verbatim — `normalizeHTTPS` collapses whitespace and nothing else. `verifySVCB` has done subset matching since #108; the sibling record type never got it, so a CDN-proxied name in a signed zone hard-fails `verify-dns` on a record the operator cannot publish.

**What.** `verifyHTTPS` now parses both sides with `parseSVCBValue` and matches with `matchesSVCBSubset`, exactly as `verifySVCB` does: priority equal, effective TargetName equal (§2.5.2), every expected SvcParam present, extras tolerated (§8).

One refinement SVCB did not need: `alpn` is a list (§7.1.1), not a scalar. `svcParamMatches` compares it as set inclusion — the expected ids must be advertised, the server may advertise more — so an expected `alpn=h2` is satisfied by a served `alpn=h3,h2`. Every other key still compares by equality. `splitAlpn` splits on unescaped commas only, since §7.1.1 lets a protocol id contain a comma escaped as `\,`.

That rule also reaches SVCB, in the same RFC-correct direction: a row carrying `alpn=a2a,mcp` now satisfies an expected `alpn=a2a` instead of hard-failing. Calling it out because it is behaviour outside the record type in the title. It is the same argument as the subset match itself — the Consolidated Approach designs for one row shared by several families, and a scalar comparison of a list-valued param defeats that.

Real tampering still fails: a different priority, a different effective target, or an `alpn` that does not carry the expected protocol.

`normalizeHTTPS` goes with its last caller.

**Tests.** `TestLookupVerifier_HTTPSSubset`, five cases: the CDN-synthesized shape (`alpn="h3,h2"` with `ech` and both hint families) matches `1 . alpn=h2`; the explicit TargetName form matches `.`; `alpn=h3` alone does not; priority `2` does not; a record missing an expected `port=` does not. Each asserts `Actual` carries the served value either way, so the operator sees what the zone answered. `TestSVCParamMatches` covers the per-key rule directly, including the escaped-comma id that no live record in the suite produces. `TestLookupVerifier_HTTPSBadExpectedValue` pins the branch that parsing the expected value introduces.

**Gate.** `go vet` clean, `go test -race ./...` clean, `make test-cover` 90.7% against the 90% floor. `verifyHTTPS` 92.6%, `matchesSVCBSubset` / `svcParamMatches` / `splitAlpn` 100%. `golangci-lint` was not run locally — not installed here; happy to fix anything CI flags.

**Not in this PR.** Whether `Required: false` should also exempt a *present but unmatched* HTTPS record in a signed zone is a lifecycle-layer decision, and this change makes it unnecessary for the CDN case rather than settling it.

### AI assistance

### DIFF
--- a/internal/adapter/dns/dns_test.go
+++ b/internal/adapter/dns/dns_test.go
@@ -292,6 +292,132 @@ func TestLookupVerifier_HTTPSMatch(t *testing.T) {
 	}
 }
 
+// TestLookupVerifier_HTTPSSubset pins the subset rule on the HTTPS RR:
+// the expected SvcParams must be present, extras are tolerated, alpn is
+// compared as a list (RFC 9460 §7.1.1), and the TargetName forms of
+// §2.5.2 designate the same target. The first case is the shape a CDN
+// edge synthesizes on a proxied name — the operator cannot publish any
+// other value, so a verbatim comparison would strand them.
+func TestLookupVerifier_HTTPSSubset(t *testing.T) {
+	t.Parallel()
+	tests := []struct {
+		name  string
+		rr    string
+		want  string
+		found bool
+	}{
+		{
+			name:  "cdn_synthesized_record_satisfies_expected_alpn",
+			rr:    `agent.example.com. 300 IN HTTPS 1 . alpn="h3,h2" ipv4hint=192.0.2.1 ech="AEX+DQBB" ipv6hint=2001:db8::1`,
+			want:  `1 . alpn=h2`,
+			found: true,
+		},
+		{
+			// §2.5.2, as fixed for SVCB in #108: at this owner name the
+			// explicit FQDN and "." are the same target.
+			name:  "explicit_targetname_matches_dot",
+			rr:    `agent.example.com. 300 IN HTTPS 1 agent.example.com. alpn="h2"`,
+			want:  `1 . alpn=h2`,
+			found: true,
+		},
+		{
+			name:  "alpn_without_the_expected_protocol_does_not_match",
+			rr:    `agent.example.com. 300 IN HTTPS 1 . alpn="h3"`,
+			want:  `1 . alpn=h2`,
+			found: false,
+		},
+		{
+			name:  "different_priority_does_not_match",
+			rr:    `agent.example.com. 300 IN HTTPS 2 . alpn="h2"`,
+			want:  `1 . alpn=h2`,
+			found: false,
+		},
+		{
+			name:  "missing_expected_param_does_not_match",
+			rr:    `agent.example.com. 300 IN HTTPS 1 . alpn="h2"`,
+			want:  `1 . alpn=h2 port=8443`,
+			found: false,
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			s := newTestServer(t)
+			s.add("agent.example.com.", "HTTPS", tc.rr)
+
+			recs := []domain.ExpectedDNSRecord{{
+				Name: "agent.example.com", Type: domain.DNSRecordHTTPS,
+				Value: tc.want, Required: false,
+			}}
+			got := s.verifyAgainst(t, recs)
+			if got[0].found != tc.found {
+				t.Errorf("found = %v, want %v (actual=%q)", got[0].found, tc.found, got[0].actual)
+			}
+			// Whether or not it matched, the served value is reported so
+			// the operator sees what the zone actually answered.
+			if got[0].actual == "" {
+				t.Error("Actual must carry the served record")
+			}
+		})
+	}
+}
+
+// TestLookupVerifier_HTTPSBadExpectedValue pins the branch that parsing
+// the expected value introduces: a stored value the RA could not have
+// written surfaces as an error, not as a silent not-found.
+func TestLookupVerifier_HTTPSBadExpectedValue(t *testing.T) {
+	t.Parallel()
+	s := newTestServer(t)
+	s.add("agent.example.com.", "HTTPS", `agent.example.com. 300 IN HTTPS 1 . alpn="h2"`)
+
+	recs := []domain.ExpectedDNSRecord{{
+		Name: "agent.example.com", Type: domain.DNSRecordHTTPS,
+		Value: "1", Required: false,
+	}}
+	got := s.verifyAgainst(t, recs)
+	if got[0].found {
+		t.Error("an unparseable expected value must not match")
+	}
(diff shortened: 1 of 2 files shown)


csnitker-godaddy (approved):
Read this against lookup.go and lifecycle.go on main and the mechanism in #115 is exactly right: the Required=false exemption only ever covered an absent record, so a proxied name in a signed zone had no way through. Routing HTTPS through the SVCB matcher is the right fix, and treating alpn as a list is the correct reading of §7.1.1. The negative cases and the Actual assertions are what I'd want. Approving.

Two small things, neither blocking. `splitAlpn` handles a single-escaped `\,`, but what it's handed comes from miekg's `String()`, which emits the zone-file double-escaped form (`\\\044` for a comma), so a comma-bearing id wouldn't round-trip. Harmless since no ANS alpn has one, but the comment describes a form that isn't what flows through. And the subset rule also tolerates served keys that change behavior, like a `port=` where the expected row has none. Same as SVCB today, so not for this PR, just noting it as a possible follow-up.

The spec still describes SvcParam matching as value-equality (ans-dnsaid §5) and says nothing for the HTTPS RR. I'll pick that up on the registry side.

PR #118, opened 2026-09-09
fix(dns): split alpn on the escape form that actually reaches the verifier

### LINKED ISSUE #117: splitAlpn documents a single-escaped alpn form that never reaches it
Raised by @csnitker-godaddy while reviewing #116, and correct — filing it so the fix has an issue to close.

## The claim in the comment

`splitAlpn` (`internal/adapter/dns/lookup.go`) documents itself as handling a comma inside a protocol id "escaped as `\,` in presentation form". That is not the form the verifier is handed.

RFC 9460 §A.1 has an alpn value escaped once for the list and then again for the zone file, and `miekg/dns` emits that double form from `SVCBAlpn.String()`. Rendering `[]string{"f,oo", "h2"}` through `formatHTTPSValue` gives:

```
1 . alpn=f\\\044oo,h2
```

The comma inside the id is `\044`; there is no comma byte in it. So every comma in the string `splitAlpn` receives is a separator, and the premise the comment is built on never occurs.

## What actually goes wrong

Nothing is mis-split — the escaped form carries no comma to split on. The defect is one layer down: the loop rebuilds each segment byte by byte and consumes one backslash on the way, so `f\\\044oo` comes back as `f\044oo`. That text matches neither the served value nor an identically written expected one. Harmless today, since no ANS `alpn` carries a comma, but it is a silent difference sitting in a comparison function.

## Fix

Return the segments verbatim as substrings instead of rebuilding them, and keep the escape-skipping only for its real case: an expected value someone writes by hand in the single-escaped form. Then say in the comment what arrives, rather than what a person would write.

A test that renders through `formatHTTPSValue` rather than hard-coding the emission would also catch it if `miekg/dns` ever changes the form under us.

PR follows.

## Not in scope

The other note on #116 — that the subset rule tolerates a served key the expected row lacks, such as a `port=` — is a separate question and applies equally to SVCB. Leaving it where you put it.

martian-pink:
Fixes #117. Follow-up to @csnitker-godaddy's review note on #116 — the call was right.

**BLUF.** `splitAlpn` said a comma inside a protocol id arrives escaped as `\,`. What arrives is the double-escaped zone-file form: `miekg/dns` renders `f,oo` as `f\\\044oo`, carrying no comma byte. Every comma the verifier splits on is a separator, so the premise the helper was written against never occurs.

**What was actually wrong.** Nothing was mis-split — there is no comma in the escaped form to split on. One layer down, though, the loop rebuilt each segment byte by byte and consumed a backslash doing it, so `f\\\044oo` came back as `f\044oo`: text matching neither the served value nor an identically written expected one. A silent difference inside a comparison function.

**What changed.** Segments are returned verbatim as substrings; the escape handling stays, now only for its real case — an expected value written by hand in the single-escaped form. The comment says what arrives instead of what a person would write.

**Tests.** `TestSplitAlpnAgainstEmittedForm` builds an `SVCBAlpn` with a comma-bearing id, renders it through `formatHTTPSValue`, and asserts on that: the id stays one segment, comes back verbatim, matches an identically written expected value, and is not readable as the separate ids it resembles. Rendering rather than hard-coding means the test fails if the emission ever changes under us. The two hand-written cases in `TestSVCParamMatches` stay and now say which form they carry.

**Gate.** `go vet` clean, `go test -race ./...` clean, `make test-cover` 90.7% against the 90% floor, `splitAlpn` and `svcParamMatches` at 100%. `golangci-lint` not run locally — not installed here.

**Not in scope.** The other note on #116, that the subset rule tolerates a served key the expected row lacks (a `port=`, say), applies equally to SVCB and is a design question rather than a bug in this helper. Left where you put it.

### AI assistance

### DIFF
--- a/internal/adapter/dns/dns_test.go
+++ b/internal/adapter/dns/dns_test.go
@@ -401,8 +401,10 @@ func TestSVCParamMatches(t *testing.T) {
 		{"alpn_list_in_longer_list", "alpn", "h2,h3", "h3,h2,http/1.1", true},
 		{"alpn_absent_from_list", "alpn", "h2", "h3,http/1.1", false},
 		{"alpn_exact", "alpn", "h2", "h2", true},
-		// A protocol id may itself contain a comma, escaped in
-		// presentation form (§7.1.1); the escape must not split the id.
+		// A protocol id may itself contain a comma. These two carry the
+		// single-escaped form a person might write by hand; the form that
+		// actually arrives from the wire is exercised in
+		// TestSplitAlpnAgainstEmittedForm below.
 		{"alpn_escaped_comma_is_one_id", "alpn", `f\,oo`, `h2,f\,oo`, true},
 		{"alpn_escaped_comma_not_two_ids", "alpn", "oo", `h2,f\,oo`, false},
 	}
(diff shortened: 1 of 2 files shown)


csnitker-godaddy (approved):
The change looks right. I checked the emitted form and both HTTPS and SVCB lookup paths. The DNS tests and race detector pass.

One correction to the explanation, including my earlier note on #116: identically escaped expected and served values already matched because both went through `splitAlpn`. The helper lost the original representation, but that alone didn’t cause a mismatch.

There’s a useful regression case this fixes, though: an ID containing `h` followed by byte `0x02` renders as `h\002`, and the old code incorrectly matches it against the literal ID `h002`. This PR correctly rejects that. I reproduced it through both record types. Worth adding that case alongside the current test, but it doesn’t block approval.

PR #125, opened 2026-09-14
test: enforce internal, domain, and crypto coverage gates

### LINKED ISSUE #121: test: enforce the documented internal, domain, and crypto coverage gates
The documented coverage requirements are not fully enforced by the existing aggregate coverage check. A passing internal total does not establish that the domain and crypto packages meet their separate requirements.

Reproduce on `83875d0572287821489c8dc5fc9cf37142ee35f0` by inspecting `make test-cover`: its gate does not independently enforce internal coverage at 90%, domain at 100%, and crypto at 95%.

Enforce all three requirements against statement-weighted `internal/` coverage, fail when required package coverage is missing, and keep command wiring outside that denominator. Cover the parser's threshold and duplicate-block behavior, and add the missing domain/crypto regressions needed to meet the gates. Document unreachable crypto branches where the package permits a defensive coverage exception.

This issue covers the validation baseline for the following ACME, TL, and renewal fixes.

csnitker-godaddy:
`make test-cover` now enforces the documented coverage requirements separately: at least 90% for `internal/`, 100% for `internal/domain`, and 95% for `internal/crypto`. Command wiring remains excluded from the denominator, missing required packages fail, and repeated observations of a coverage block are counted once.

Added domain and crypto regressions and documented the defensive crypto branches that cannot be reached through the preceding guards. Synthetic coverage profiles verify each threshold, exclusions, missing packages, and duplicate observations.

Validation: `make build`, `make check`, and `make test-race` pass on this branch. Coverage is internal **91.01%**, domain **100.00%**, crypto **98.45%**.

First PR in the deployment-fix stack. No public API or wire-format changes.

Fixes #121.

### AI assistance


scripts/check-coverage.awk
```diff
@@ -0,0 +1,36 @@
+# Count each source block once across cross-package coverage profiles.
+NR == 1 { next }
+{
+    split($1, location, ":")
+    file[$1] = location[1]
+    statements[$1] = $2
+    if ($3 > 0) hit[$1] = 1
+}
+END {
```
kperry-godaddy, line 9: Not a blocker: `minimum` is never validated, so an empty, unset or non-numeric value coerces to 0 in the comparison at line 30 and the internal floor passes at any coverage while still printing "Coverage thresholds passed." Today the Makefile always supplies it, so the trigger is `make test-cover COVERAGE_THRESHOLD=` or a later edit that drops the `-v`; failing closed costs four lines:

```suggestion
END {
    if (minimum !~ /^[0-9]+(\.[0-9]+)?$/ || minimum + 0 <= 0) {
        print "FAIL: minimum must be a positive percentage (pass -v minimum=N)"
        exit 1
    }
```

A row in `TestCoverageGate` that omits `-v minimum` and expects exit 1 pins it.

kperry-godaddy (approved):
Approving. The awk gate does what #121 asks: statement-weighted floors over `internal/`, once-per-block union across the cross-package profile, and a hard failure when domain or crypto is missing from the profile. The motivation holds up on the numbers: the base passes the old aggregate check at 90.7% total while domain sits at 99.84% and crypto at 94.14%, and the new gate fails that profile and passes this branch at 91.02 / 100.00 / 98.45. All nine SAFETY/NOTE annotations trace to the guard they cite.

Nothing here blocks. Three things worth tidying while the files are open: one inline with a suggestion, two below because they sit outside the diff.

### Outside this diff
- `Makefile:83` — not a blocker: `grep -v -e '/scripts/'` no longer excludes the new test-only root package `github.com/agentnameservice/ans/scripts` (the import path has no trailing slash), so it now lands in `-coverpkg`, `go vet` and lint, and the comment at 71-82 describes an exclusion that no longer happens. `-e '/scripts'` matches both the package and its children; the comment could also name the three floors instead of "the 90% gate".
- `internal/domain/agent_test.go:277-291` — not a blocker: `TestNewRegistration_InvalidEndpoint` passes an empty displayName, trips `MISSING_DISPLAY_NAME`, and never reaches `eps.Validate`, which is why that block was uncovered before this PR. The new table row at 95-97 is the real test; deleting the old one (or giving it a displayName and asserting `ENDPOINT_HOST_MISMATCH`) removes a test that claims coverage it does not provide.

While we're in here: the internal floor comes from `COVERAGE_THRESHOLD` while 100 and 95 are literals in the awk, repeated in the FAIL text and in the test's expected string. Either pass all three from the Makefile or write the 90 beside the 95 so there is one place to read. The single "90% gate" wording in README.md, CLAUDE.md:212 and ci.yml can follow in the same change or a named follow-up.

PR #126, opened 2026-09-14
fix(ra): bind ACME issuance to owners and verify certificate chains

### LINKED ISSUE #122: ra: bind ACME authorization to its owner and enforce server certificate trust
The RA can treat an ACME provider's reusable authorization as proof from the current authenticated owner. The executable also constructs its server-certificate validator with chain verification disabled, so possession of a matching public certificate or an arbitrary private chain can be confused with an authorized, trusted registration.

- Return an already-authorized provider order with no challenges, then call registration or renewal `verify-acme` with no owner-published artifact. The RA bypasses its challenge gate.
- Reuse a provider's pending authorization across two RA owners on the shared account. The challenge key authorization is shared.
- Present a matching server certificate signed by an unrelated private CA to the executable's validator. The configured chain-verification bypass accepts it.

Require current owner proof for registration and renewal, isolate persistent ACME accounts by authenticated RA owner, and bind finalization to that owner. Retry may reuse proof already recorded by the RA. Legacy pending orders without owner-bound proof must fail closed.

Enable chain verification in executable wiring. Trust system roots, explicitly configured private/staging roots, and the explicitly selected local server issuer. Update demo flows to use the configured issuer or trusted BYOC input.

Keep the current HTTP request/response fields. The cached-authorization exception in the registry prose needs corresponding clarification.

csnitker-godaddy:
Registration and renewal now require domain-control proof from the authenticated owner, including when the ACME provider reports an already-authorized order. Authorization belonging to another owner cannot be reused. Retrying an order whose proof the RA has already verified remains supported.

Server certificates must chain to system roots, an explicitly configured root, or the selected local issuer. The demo defaults to CSR issuance; BYOC requires a trusted certificate. Staging and private roots can be configured explicitly.

### Contract comparison

The existing request shape in [`spec/api-spec-v2.yaml`](https://github.com/agentnameservice/ans/blob/main/spec/api-spec-v2.yaml) is unchanged:

```yaml
ServerCertificateRenewalRequest:
  type: object
  properties:
    serverCsrPEM:
      type: string
    serverCertificatePEM:
      type: string
    serverCertificateChainPEM:
      type: string
  oneOf:
    - required: [serverCsrPEM]
    - required: [serverCertificatePEM]
```

`RenewalSubmissionResponse` already defines `status: [PENDING_VALIDATION, ISSUING_CERTIFICATE]` and nullable `challenges.dns01` / `challenges.http01`. Fresh owner proof uses those existing fields. No HTTP fields, routes, enums, envelope fields, or signature formats change.

The behavior intentionally tightens authorization reuse. Registry ANS-1 prose that permits an already-authorized order to bypass owner proof needs a matching clarification; this PR does not claim that prose is already updated. Existing pending legacy orders without persisted proof require a new registration/version or renewal rather than silent approval.

### Validation

Tests cover cached authorization across owners, pending authorization, restart and retry behavior, and trusted and untrusted certificate chains.

`make build`, `make check`, and `make test-race` pass on this branch. Coverage is internal **91.01%**, domain **100.00%**, crypto **98.45%**.

Second PR in the deployment-fix stack. Preserve the RA and CA data when upgrading.

Fixes #122.

Depends on #125; this PR targets `fix/coverage-gates` to keep its review diff scoped.

### AI assistance


internal/adapter/cert/acme_owner.go
```diff
@@ -0,0 +1,93 @@
...
+	if issuer, ok := a.owners.Get(accountID); ok {
+		return issuer, nil
+	}
+	issuer, err := NewACMEIssuer(a.directoryURL, a.email, filepath.Join(a.dataDir, "owners", accountID), a.options...)
+	if err != nil {
+		return nil, fmt.Errorf("cert: load owner ACME account: %w", err)
+	}
+	a.owners.Add(accountID, issuer)
+	return issuer, nil
+}
```
kperry-godaddy: `ownerIssuer` creates a fresh ACME account for every first-time owner, and the child's first `CreateOrder` registers it inside the registration or renewal request. Against Let's Encrypt production that runs into the limit of 10 new accounts per IP address per 3 hours, which has no override (https://letsencrypt.org/docs/rate-limits/): the eleventh new owner in a window gets a generic 500 `CERT_ORDER_FAILED` with no Retry-After, no distinct error, and no log line naming the owner. Existing owners are unaffected because the key is persisted, so this is an onboarding ceiling, and one that self-serve identities can exhaust deliberately.

Worth fixing before this ships, and it starts with a decision rather than code. If per-owner accounts stay (they do give each owner its own 300-orders-per-3h budget), make the limit first-class: detect the ACME 429/503 in `ensureRegistered`, return a retryable sentinel the service maps to 503 with Retry-After, log at WARN with the owner-account prefix, and document the ceiling next to the acme config block. The alternative is one provider account plus a mandatory RA-issued, owner-bound challenge on every order, which removes the accounts, the LRU and the scoped-ref format at the cost of two records for pending orders.

```go
var ae *acme.Error
if errors.As(err, &ae) && (ae.StatusCode == http.StatusTooManyRequests || ae.StatusCode == http.StatusServiceUnavailable) {
	a.logger.Warn().Err(err).Str("retryAfter", ae.Header.Get("Retry-After")).Msg("acme account registration throttled by provider")
	return fmt.Errorf("cert: acme account registration: %w", port.ErrProviderThrottled)
}
```

internal/ra/service/order_proof.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/agentnameservice/ans/internal/domain"
+	"github.com/agentnameservice/ans/internal/port"
+)
+
+func (s *RegistrationService) createServerOrder(ctx context.Context, ownerID, fqdn string) (*domain.CertificateOrder, error) {
+	if scoped, ok := s.serverCA.(port.OwnerScopedServerCertificateIssuer); ok {
+		return scoped.CreateOrderForOwner(ctx, ownerID, fqdn)
+	}
+	return s.serverCA.CreateOrder(ctx, fqdn)
+}
```
kperry-godaddy, line 23: Owner isolation is discovered here by a two-value type assertion, with the shared-account `CreateOrder` as the silent fallback. The finalize side of the same change took the opposite shape: `OwnerID` is a plain field on `FinalizeOrderRequest`. Nothing pins `*cert.ACMEIssuer` to `OwnerScopedServerCertificateIssuer` at compile time, all five test fakes take the fallback, and any decorator around the issuer (logging, metrics, retry) routes every order through the shared account. For the shipped adapter that becomes a 100% verify-acme outage via the finalize guard rather than a bypass; a future account-caching adapter that omits the method reopens the second reproduction in #122 with no error and no log.

Worth fixing before this ships, because #127 and #128 build on this port. Put the owner on the port itself, mirroring `FinalizeOrderRequest`, and delete the side interface and the probe:

```go
type CreateOrderRequest struct {
	OwnerID string
	FQDN    string
}

type ServerCertificateIssuer interface {
	CreateOrder(ctx context.Context, req CreateOrderRequest) (*domain.CertificateOrder, error)
	FinalizeOrder(ctx context.Context, req FinalizeOrderRequest) (*IssuedCert, error)
	GetCACertificate(ctx context.Context) (string, error)
}
```

`ServerSelfCA` ignores the owner (its tokens are per-order random), `ACMEIssuer.CreateOrder` takes over the owner-scoped body with an unexported child helper so the per-owner child does not re-dispatch, and the compiler lists the call sites: the service fakes in `order_flow_test.go`, `acme_test.go`, `cert_test.go`, `acme_owner_test.go`, `le_live_smoke_test.go` and `cmd/ans-ra/validator_test.go`. If that has to wait, `var _ port.OwnerScopedServerCertificateIssuer = (*ACMEIssuer)(nil)` plus one service test asserting the scoped method is preferred is the minimum pin.

internal/adapter/cert/acme_owner.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+	root, err := owner.GetCACertificate(ctx)
+	if err != nil {
+		return nil, err
+	}
+	a.mu.Lock()
+	a.chainRootPEM = root
+	a.mu.Unlock()
+	return issued, nil
+}
```
kperry-godaddy: `ACMEIssuer` is now both a single-account ACME client and a registry and factory of itself: six registry fields, children of the same type (which can spawn `owners/<id>/owners/<id>`), `FinalizeOrder` dispatching on the `ans-acme-owner:` prefix, and this function clearing `req.OwnerID` so the child skips the owner check. That reset is the only reason the `req.OwnerID != ""` condition at `acme.go:277` exists, and it leaves a fail-open branch: a scoped ref with no owner finalizes on whatever account the ref names. Both RA callers set `OwnerID` today, so it is unreachable through HTTP, but it rests on caller discipline at a trust boundary, and every later change to finalize semantics has to reason about two modes distinguished by string content. The parent's own account key is also still generated at boot and never used.

Worth fixing before this ships, and it pairs with the port change above. Extract the registry into a wrapper that implements the port, performs the owner check once, strips the prefix and delegates to the child's plain methods; `ACMEIssuer` goes back to a single-account client with no prefix dispatch and no `OwnerID` reset, and `newAccount` becomes the seam for eviction and reload tests:

```go
type ownerScopedACMEIssuer struct {
	directoryURL, email, dataDir string
	opts       []ACMEIssuerOption
	logger     zerolog.Logger
	mu         sync.Mutex
	accounts   *lru.Cache[string, *ACMEIssuer]
	newAccount func(dataDir string) (*ACMEIssuer, error)
}
```

Whatever shape wins, require a non-empty `OwnerID` for any ref carrying the scoped prefix. One smaller thing in the same function: `owner.GetCACertificate` failing after a successful finalize turns an issued certificate into a permanent 500 when the provider returns a leaf-only chain (RFC 8555 §9.1 allows it), and nothing reads the copied root in ACME mode. Making the copy best-effort or dropping it avoids that.

kperry-godaddy (changes requested):
Owner-bound ACME issuance and always-on chain verification close all three problems in #122 for the shipped wiring, and the removed code confirms each of them was real. What shapes my read is how owner isolation is delivered: per-owner Let's Encrypt accounts, reached through an optional interface. Both work today; both carry costs that are cheaper to settle now than after #127 and #128 build on this port shape.

Three things I'd fix before this ships, all inline. Three non-blocking notes below.

- Logging: the new files add no log statements. Gate denials, the owner-mismatch rejection, per-owner account creation and the substitution of RA challenges for a born-ready provider order are all silent, and the handler logs only >= 500, so the 422s this change introduces leave no trace in RA logs. `gateOrderChallenges` also still logs through the package-global logger rather than `s.logger`.
- Upgrade path: an ACME order pending from before this change passes the gate after the owner publishes the record, then fails at finalize with a 422 that blames the provider and marks the order FAILED terminally; a born-ready ISSUING row is told to publish from an empty `challenges[]`. An explicit case for `len(order.Challenges) == 0` with terminal guidance, a distinct sentinel from the adapter so the service does not `MarkFailed` on it, and a one-paragraph upgrade note (drain pending ACME orders first) would spare the first operator to hit it.
- ANS-1 §7 and conformance item 2 in ans-registry still describe the authorization-reuse exception this PR removes. A tracking issue there keeps the reference implementation and the normative text from drifting apart.

PR #131, opened 2026-09-18
feat(ans-verify): accept an FQDN and resolve its ANS badge

### LINKED ISSUE #120: ans-verify: accept an FQDN and resolve its ANS badge
Track the FQDN entry point discussed in #99 separately from the proposal for non-ANS verification.

Add an explicit FQDN input while preserving existing agent-ID usage. Resolve `_ans-badge.<fqdn>` and use the discovered ANS registration information as input to the existing receipt-verification flow.

Acceptance criteria:

- Existing agent-ID input remains supported.
- FQDN input resolves the ANS badge and runs the existing receipt and proof checks with the current trust configuration.
- Missing or malformed badge records and DNS lookup errors produce clear failures and a nonzero exit status.
- Existing metadata-hash verification options remain available.
- There is no fallback to SVCB endpoint discovery, non-ANS card verification, or a reachable-but-unregistered result.

This changes how an ANS registration is located; it does not change the evidence required to verify it.

Related: #99 and #112.

sparkmastergrape:
Closes #120 — the FQDN entry point split out of #99, and the "happy to do it next as its own PR" from #112.

**What.** `ans-verify` gains `-fqdn`. It resolves TXT at `_ans-badge.<fqdn>`, requires `v=ans-badge1`, and reads the agent id out of the `url=` value's `/v1/agents/<id>` path, which is the shape `BadgeRecord` writes. That id then enters the existing receipt and proof flow with nothing else altered. `-agent` and the positional agent id keep working unchanged; passing both is an error. `-check-metadata` and `-metadata-timeout` apply on the FQDN path too.

This changes how a registration is located, not the evidence required to verify it. No SVCB endpoint discovery, no agent-card fetch, no reachable-but-unregistered outcome.

**The log the badge names.** The badge is published by the same operator as the agent, so it does not get to move verification to a log of its choosing. An explicit `-url` wins, and a badge that disagrees is reported (`⚠ badge names X; verifying against -url Y instead`). The badge's log is adopted only when `-url` was left at its default, where the operator expressed no preference and the `localhost:18081` default would be useless. Say the word if you would rather it never adopt the badge's log and always require an explicit `-url`; that is a one-line change.

**Failures, all exit 1 with a specific message.** Lookup error, NXDOMAIN, no `ans-badge1` record at the name, a name publishing two different badges (which registration to verify is ambiguous), or a `url=` that is not a `/v1/agents/<agentId>` path. That last one also rejects the endpoint-URL form `BadgeRecord` falls back to when a deployment configures no TL base URL: it names a reachable agent, not a log, so there is no ANS evidence to check. The agent id must match the UUID shape the RA issues before it is interpolated into a fetch path, reusing the walker's `agentIDPattern` for the same reason it exists there.

**`-dns host:port`** aims the lookup at one resolver, which the local `ans-dns` dev server needs. A failed lookup names that resolver: Go's own resolver error reports the server from the system configuration even when `Dial` sent the query elsewhere, which reads as though `-dns` was ignored.

**Tests.** `badge_test.go`: the canonical record, five tolerated forms (trailing slash, no spaces, a log under a path prefix, `http` for local, uppercase keys), nine rejected ones, the badge owner name actually queried, unrelated TXT at the same name ignored, the same badge published twice accepted, seven resolve failures, the three log-selection branches, `flagWasSet`, and `normalizeResolverAddr`. Two of them use a real nameserver: one serving the badge over UDP through `newTXTLookup`, one pointed at a dead port to check the error names the resolver.

**Gate.** gofmt clean, `go vet` ok, `golangci-lint` (v2.11.4) 0 issues, `go test -race` clean, every new function 87–100% covered. Also smoke-tested end to end with the real binaries: `ans-dns serve` publishing the badge, then `ans-verify -fqdn` resolving it and proceeding into Step 1; plus the `-url` disagreement, the no-badge, the both-flags and the no-target paths.


cmd/ans-verify/badge.go
```diff
@@ -0,0 +1,346 @@
...
+	fmt.Fprintln(out, "── Step 0: Resolve ANS badge ──")
+	fmt.Fprintf(out, "  ✓ %s TXT: %s\n", b.Owner, b.Raw)
+	if b.Version != "" {
+		fmt.Fprintf(out, "    version: %s\n", b.Version)
+	}
+
+	baseURL := b.TLBaseURL
+	switch {
+	case !in.URLExplicit:
+		fmt.Fprintf(out, "    log:     %s (named by the badge; -url not set)\n", baseURL)
```
csnitker-godaddy, line 322: [Nonblocking] Explain the trust limits of automatic log discovery

The existing CLI fetches `/root-keys` from its configured log unless `-pubkey` is supplied. This branch additionally lets the target's DNS select that log. That is reasonable for receipt inspection, but success establishes cryptographic consistency with the discovered log's advertised keys; it does not establish that the log is independently trusted. Please make that limitation explicit in the output/docs and point users who want to select their verification authority independently to `-url` or a pinned `-pubkey`.

cmd/ans-verify/main.go
```diff
@@ -89,19 +109,33 @@ func main() {
...
 	fmt.Println("=== ANS SCITT Receipt Verifier ===")
+	if fqdn != "" {
+		agentID, baseURL = badgeStep(context.Background(), badgeStepInput{
+			FQDN:          fqdn,
+			Resolver:      dnsServer,
+			Timeout:       dnsTimeout,
+			ConfiguredURL: baseURL,
+			URLExplicit:   urlExplicit,
+			Out:           os.Stdout,
+		})
```
csnitker-godaddy, line 137: [P1] Bind the verified registration back to the requested FQDN

After this resolution the FQDN is never checked against the signed receipt. Even with an explicit `-url` and independently pinned `-pubkey`, I could point `_ans-badge.unregistered.example` at a valid registration for `registered.example`; both the receipt and fresh status token named `registered.example`, but the CLI printed `VERIFIED` and exited 0. An unregistered hostname can therefore borrow another registration's proof without breaking a signature. Carry the normalized requested hostname into the verification step and reject when the signed registration's primary host/ANSName does not agree.

cmd/ans-verify/badge.go
```diff
@@ -0,0 +1,346 @@
...
+// name with no resolvable badge has no ANS registration this tool can
+// locate, and falling back to anything else would be the non-ANS
+// verification path this command does not have.
+func resolveBadge(ctx context.Context, lookup txtLookupFunc, fqdn string) (badge, error) {
+	name := strings.TrimSuffix(strings.TrimSpace(fqdn), ".")
+	if name == "" {
+		return badge{}, errors.New("empty FQDN")
+	}
+	owner := badgeOwnerPrefix + name
+	txts, err := lookup(ctx, owner)
```
csnitker-godaddy, line 125: [P2] Query the badge owner as an absolute DNS name

The trailing dot is stripped from the supplied FQDN, but the lookup name is not rooted again. On systems with a DNS search list, even `-fqdn absolute.example.` can fall back to `_ans-badge.absolute.example.<search-domain>.`. I reproduced this with no TXT at `_ans-badge.absolute.example.` and a badge only at the search-qualified name: the CLI reported that it found the original owner's badge and exited 0. Append the DNS root dot to the lookup owner after normalization so another namespace cannot satisfy a missing FQDN badge.

csnitker-godaddy (changes requested):
The FQDN-only scope matches #120. Two issues need correction before merge: the signed registration is not checked against the requested hostname, and the lookup can fall through DNS search domains even for an explicitly absolute input.

The focused race tests, go vet, and formatting checks pass. I also exercised the built CLI with signed receipt/status fixtures and loopback DNS/HTTP servers. Existing explicit and positional agent IDs, wrong-key rejection, metadata-mismatch rejection, the metadata opt-out, and the ordinary no-badge failure behaved as expected. Both incorrect-hostname cases still returned success. The separate comment about DNS-selected log trust is nonblocking for this diagnostic receipt verifier.

PR #135, opened 2026-09-21
fix(docs): make Swagger work on the deployed service origin

### LINKED ISSUE #132: Swagger UI fails to parse RA spec and targets incorrect service URLs
The RA OpenAPI document defines `components.schemas.Problem` twice, causing Swagger UI to reject it with a duplicated mapping key error. The RA and TL OpenAPI server URLs also direct interactive requests to localhost or example hosts when the services run on a public domain.

Remove the duplicate mapping, use service-origin URLs, and synchronize the embedded documents with the canonical specs.

csnitker-godaddy:
The RA Swagger UI cannot parse its OpenAPI document because `components.schemas.Problem` is defined twice. Both services also direct interactive requests to localhost or example hosts instead of the deployed origin.

Remove the duplicate RA schema and use same-origin server URLs: `/v2` for RA and `/` for TL. Keep the canonical documents and their embedded copies identical. Handler responses are unchanged.

Validation: duplicate-key-aware YAML parsing, canonical/embedded byte comparison, and whitespace checks.

Fixes #132.

### AI assistance


spec/api-spec-v2.yaml
```diff
@@ -43,12 +43,8 @@ info:
     url: https://www.apache.org/licenses/LICENSE-2.0.html
 
 servers:
-  - url: http://localhost:18080/v2
-    description: Local dev — quickstart ra-local.yaml
-  - url: https://api.ans.example.org/v2
-    description: Main ANS API server (v2)
-  - url: https://test.ans.example.org/v2
-    description: Test server (v2)
+  - url: /v2
```
kperry-godaddy, line 46: The root list is fixed, but the operation-level override on `/v1/agents/events` (lines 96-100) still carries `http://localhost:18080` and `https://api.ans.example.org`. A path item's `servers` replaces the root list for that path (OAS 3.0.3, Path Item Object: https://spec.openapis.org/oas/v3.0.3#path-item-object), so on a deployed RA the events-feed "Try it out" still goes to localhost, which is the behaviour #132 describes. Same treatment there, then re-copy the embedded `ra.yaml`:

```yaml
  /v1/agents/events:
    servers:
      - url: /
        description: This Registration Authority (the feed is served at the origin root, not under /v2)
```

Worth fixing before this ships since it is the same defect on the one route that lives outside `/v2`; the change is four lines and the docsui test will catch a missed copy.

kperry-godaddy (changes requested):
The relative `servers` entries are the right call, and I checked the mechanics rather than trusting them: swagger-js resolves a relative server URL against the document URL and strips the trailing slash before it appends the operation path, so the TL's `url: /` produces clean `https://host/v1/...` requests. Removing the second `Problem` also kept the accurate copy: the generic 500 path in `errors.go` emits no `code`, so the deleted definition's `required: code` was the wrong one. Both documents now pass a duplicate-key-strict YAML parse, and the embedded copies are byte-identical to `spec/`.

One thing I'd settle before this merges, inline on the `servers` block. Two small notes below.

- 97 `$ref`s still point at `ErrorResponse` and 15 at `Problem`, while every RA failure path emits Problem Details. The deleted duplicate carried the only note that this migration is tracked separately, and I couldn't find an issue for it. Opening one would keep the work item from disappearing with the comment.
- Dropping the named hosts changes what a code generator seeds as its default base URL from `servers[0].url`. If any SDK build reads this document directly, it will want an explicit base URL now.

Issue #134, opened 2026-09-21
Deduplicate the RA certificate issuance phase literal


csnitker-godaddy:
The RA status DTO repeats `CERTIFICATE_ISSUANCE` in its phase and completed/pending step mappings. Extract a shared constant to satisfy the repeated-string lint rule while preserving the response values.

PR #137, opened 2026-09-21
refactor(ra): share certificate issuance phase constant

csnitker-godaddy:
Extract the repeated `CERTIFICATE_ISSUANCE` literal into one constant shared by the RA phase and completed/pending step mappings. This satisfies the repeated-string lint rule without changing response values or fields.

Depends on #128; this PR targets `fix/renewal-publication` to keep the diff limited to the cleanup.

Fixes #134.

### AI assistance


internal/ra/handler/dto.go
```diff
@@ -459,6 +459,8 @@ func mapCSRStatus(c *domain.AgentCSR) csrStatusResponse {
 
 // ----- AgentStatus (matches V2 spec §1133) -----
 
+const phaseCertificateIssuance = "CERTIFICATE_ISSUANCE"
```
kperry-godaddy, line 462: Two things on this line. First, the premise: with the repo's `.golangci.yml` and the pinned golangci-lint v2.11.4, `./internal/ra/handler/...` reports zero issues on the base as well as here, so no repeated-string rule was failing before this change. If a different invocation produced the goconst report behind #134, it would be worth recording that command in the PR. Second, if the phase vocabulary is worth naming, it is worth naming once: `DOMAIN_VALIDATION` still appears five times in these three functions, `DNS_PROVISIONING` three times, `INITIALIZATION` once, and `COMPLETED` is borrowed from `renewalStatusCompleted`, a renewal-status constant standing in for a phase. These five strings are the `AgentStatus.phase` enum at `spec/api-spec-v2.yaml:2069`:

```go
// AgentStatus phase and step vocabulary (spec/api-spec-v2.yaml, AgentStatus.phase enum).
const (
	phaseInitialization      = "INITIALIZATION"
	phaseDomainValidation    = "DOMAIN_VALIDATION"
	phaseCertificateIssuance = "CERTIFICATE_ISSUANCE"
	phaseDNSProvisioning     = "DNS_PROVISIONING"
	phaseCompleted           = "COMPLETED"
)
```

with `phaseFor`, `completedStepsFor` and `pendingStepsFor` switched over. Either shape is fine with me once the base is sorted; `dto_more_test.go` keeps pinning the wire values.

kperry-godaddy (changes requested):
The change is harmless and the wire values don't move, so there is nothing to worry about in the diff itself. What shapes my read is where the PR is aimed. `internal/ra/handler/dto.go` is byte-identical on `main` and on `fix/renewal-publication`, so this doesn't depend on #128: it is queued behind a stack that has changes requested on it for no reason, "Fixes #134" can't auto-close from a non-default base (https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue), and if #128 is squash-merged with its branch deleted, GitHub retargets this PR to `main` with #128's commits in its diff.

I'd settle that before anything else:

```sh
git fetch origin
git rebase --onto origin/main origin/fix/renewal-publication fix/ra-certificate-phase-constant
git push --force-with-lease
gh pr edit 137 --base main
```

One note inline on the constant itself, about what the linter actually reports and how far the extraction goes.

